PR #136, opened 2024-11-25
Update PVM host functions to 0.5.0

pantrif:
I updated the accumulate and general host functions to align with the 0.5.0 changes. The transition of registers and host result codes to `uint64` will be addressed in a separate PR.


carlos-romano (approved):
Looks good to me

PR #147, opened 2024-11-27
State serialization to 0.5.0

carlos-romano:


internal/service/service.go
```diff
@@ -82,6 +83,18 @@ type PreImageMetaKey struct {
 }
 type PreimageHistoricalTimeslots []jamtime.Timeslot // Metadata for preimages (l) - TODO: There is a MaxHistoricalTimeslotsForPreimageMeta.
 
+func (p PreimageHistoricalTimeslots) ToByteSlice() []byte {
+	// Allocate a byte slice large enough to hold all Timeslot values
+	output := make([]byte, len(p)*4)
+
+	// Convert each Timeslot to bytes and copy into the output slice
+	for i, ts := range p {
```
pantrif, line 91: I think you should use jam codec for encoding timeslot

internal/state/merkle/serialization_utils_test.go
```diff
@@ -10,25 +12,63 @@ import (
...
-	binary.BigEndian.PutUint32(expectedEncodedServiceId, uint32(serviceId))
-	assert.Equal(t, expectedEncodedServiceId, stateKey[1:5])
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Generate the state key
+			stateKey := generateStateKeyInterleavedBasic(tt.i, tt.serviceId)
+
+			// Convert serviceId to bytes for verification
+			serviceIdBytes := make([]byte, 4)
+			binary.BigEndian.PutUint32(serviceIdBytes, uint32(tt.serviceId))
```
pantrif, line 46: Is there any reason why you use big endian encoding? I think you should use jam integer encoding here (little endian)

carlos-romano: If you see the implementation of generateStateKeyInterleavedBasic, we need that ordering. With Little Endian won't pass

danielvladco (dismissed):

greywolve (dismissed):

internal/state/merkle/serialization_utils_test.go
carlos-romano: Agreed that the implmentation should be different. Updated

pantrif (approved):

pantrif (approved):

PR #149, opened 2024-11-28
Fix disputes and add test vectors for v0.4.5

bamzedev:


tests/integration/disputes_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+	for _, file := range files {
+		if !strings.HasSuffix(file.Name(), ".json") {
+			continue
+		}
+
+		t.Run(file.Name(), func(t *testing.T) {
+			filePath := fmt.Sprintf("vectors/disputes/tiny/%s", file.Name())
+			data, err := ReadJSONFile(filePath)
+			require.NoError(t, err, "failed to read JSON file: %s", filePath)
```
greywolve, line 273: Only nitpick with this is that it would be nice to also see the test description somewhere.

bamzedev: Agree, but didn't want to hardcode the files. I plan on open another one for the latest version of the GP, might add it as just a comment.

PR #183, opened 2024-12-12
Add import host function

pantrif:
Import `ΩY` PVM host function

### DIFF
--- a/internal/polkavm/host_call/refine_functions.go
+++ b/internal/polkavm/host_call/refine_functions.go
@@ -4,13 +4,14 @@ import (
 	"math"
 
 	"github.com/eigerco/strawberry/internal/block"
+	"github.com/eigerco/strawberry/internal/common"
 	"github.com/eigerco/strawberry/internal/crypto"
 	"github.com/eigerco/strawberry/internal/jamtime"
 	. "github.com/eigerco/strawberry/internal/polkavm"
 	"github.com/eigerco/strawberry/internal/service"
 )
 
-// HistoricalLookup ΩH(ϱ, ω, µ, (m, e), s,d, t)
+// HistoricalLookup ΩH(ϱ, ω, µ, (m, e), s, d, t)
 func HistoricalLookup(
 	gas Gas,
 	regs Registers,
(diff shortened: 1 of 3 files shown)


internal/polkavm/host_call/refine_functions.go
```diff
@@ -70,3 +71,42 @@ func HistoricalLookup(
...
+	importedSegments []Segment,
+) (Gas, Registers, Memory, RefineContextPair, error) {
+	if gas < ImportCost {
+		return gas, regs, mem, ctxPair, ErrOutOfGas
+	}
+	gas -= ImportCost
+
+	omega7 := regs[A0]
+	omega8 := regs[A1]
+	omega9 := regs[A2]
```
danielvladco, line 90: is it possible to give more intuitive names? you can take a look at the accumulation host calls for example

internal/polkavm/host_call/refine_functions.go
```diff
@@ -70,3 +71,42 @@ func HistoricalLookup(
...
+
+	l := min(omega9, common.SizeOfSegment)
+
+	segmentToWrite := v[:l]
+	if err := mem.Write(uint32(omega8), segmentToWrite); err != nil {
+		regs[A0] = uint64(OOB)
+		return gas, regs, mem, ctxPair, nil
+	}
+
+	regs[A0] = uint64(OK)
```
danielvladco, line 110: we can use util `withCode` to keep the code more readable

internal/polkavm/host_call/refine_functions_test.go
```diff
@@ -104,3 +105,66 @@ func TestHistoricalLookup(t *testing.T) {
 	expectedGasRemaining := polkavm.Gas(initialGas) - host_call.HistoricalLookupCost - polkavm.GasCosts[polkavm.Ecalli] - polkavm.GasCosts[polkavm.JumpIndirect]
 	assert.Equal(t, expectedGasRemaining, gasRemaining)
 }
+
+func TestImport(t *testing.T) {
```
danielvladco, line 109: For now there is one test so this is good but when there is going to be other refine host calls we might want to write tests like in accumulate, where we have a bunch of testcases and only allocate registers/memory, this way I think is more readable and less boilerplate

internal/polkavm/host_call/refine_functions.go
pantrif: done

internal/polkavm/host_call/refine_functions.go
pantrif: done

danielvladco (approved):

Issue #187, opened 2024-12-17
Update codec test vectors to latest


pantrif:
The [codec](https://github.com/w3f/jamtestvectors/tree/master/codec) test vectors have been updated to reflect changes from GP v0.5.2. We need to update our repository with the latest changes and ensure that all tests pass.

PR #188, opened 2024-12-18
Update codec test vectors

pantrif:
closes https://github.com/eigerco/strawberry/issues/187


internal/work/package.go
```diff
@@ -21,10 +21,10 @@ type Package struct {
 
 // ValidateNumberOfEntries (14.4 v0.5.2)
 func (wp *Package) ValidateNumberOfEntries() error {
-	var totalExported, totalImported uint
+	var totalExported, totalImported uint16
```
greywolve, line 24: Is this change for efficiency?

greywolve (approved):

bamzedev (approved):

internal/work/package.go
pantrif: In the graypaper (GP), `N` is defined as the set of non-negative integers with **no fixed size**. However, I changed `ExportedSegments` to `uint16` to align with the ASN.1 types defined [here](https://github.com/w3f/jamtestvectors/blob/master/jam-types-asn/jam-types.asn#L143).

PR #189, opened 2024-12-18
feat: statistics tests

danielvladco:


internal/statetransition/state_transition.go
```diff
@@ -2142,13 +2146,26 @@ func calculateNewValidatorStatistics(block block.Block, timeslot jamtime.Timeslo
...
-			for _, credential := range guarantee.Credentials {
-				if credential.ValidatorIndex == v {
-					newStats[1][v].NumOfGuaranteedReports++
-				}
+		for _, msg := range reporters {
+			if slices.Equal(currValidators[v].Ed25519, msg) {
+				newStats[1][v].NumOfGuaranteedReports++
 			}
 		}
+		//for _, guarantee := range block.Extrinsic.EG.Guarantees {
```
bamzedev, line 2154: Forgot do delete?

internal/statetransition/state_transition.go
```diff
@@ -1168,8 +1168,8 @@ func calculateNewCoreAssignments(
 	validatorState validator.ValidatorState,
 	newTimeslot jamtime.Timeslot,
 	entropyPool state.EntropyPool,
-) state.CoreAssignments {
-	newAssignments := intermediateAssignments
+) (newAssignments state.CoreAssignments, reporters []ed25519.PublicKey) {
```
bamzedev, line 1171: Should we use a map instead to assure uniqueness?

bamzedev: Which is exactly what the GP says, I think. `We note that the Ed25519 key of each validator whose
signature is in a credential is placed in the reporters set R`

danielvladco: it's also hard to do because we use a slice instead of array and it is not comparable, but I can refactor that

bamzedev (approved):

Issue #141, opened 2024-11-26
Erasure coding test vectors


bamzedev:
https://github.com/w3f/jamtestvectors/pull/4

greywolve:
Moving this to the backlog until the erasure coding test vectors are ready, the current ones are incorrect.

PR #190, opened 2024-12-18
add reed solomon encoding

greywolve:
Add reed solomon encoding by wrapping the rust reed-solomon-simd crate and exposing it via FFI.

Part of #141


erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+pub unsafe extern "C" fn reed_solomon_encode(
+    original_shards_count: size_t,
+    recovery_shards_count: size_t,
+    shard_size: size_t,
+    original_shards: *const u8,
+    recovery_shards_out: *mut u8,
+) -> c_int {
+    let original_shards: Vec<_> = slice::from_raw_parts(
+        original_shards,
+        original_shards_count as usize * shard_size as usize,
```
asmie, line 23: Are we sure there won't be overflow?
If there might be, let's use, e.g. `checked_mul()` to handle that.

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn reed_solomon_encode(
+    original_shards_count: size_t,
+    recovery_shards_count: size_t,
+    shard_size: size_t,
+    original_shards: *const u8,
+    recovery_shards_out: *mut u8,
+) -> c_int {
+    let original_shards: Vec<_> = slice::from_raw_parts(
+        original_shards,
```
asmie, line 22: Should we check if that pointer is not null?

```
if original_shards.is_null() {
        return -1; // Return error code if null
    }
```

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    .collect();
+
+    match reed_solomon_simd::encode(
+        original_shards_count as usize,
+        recovery_shards_count as usize,
+        original_shards,
+    ) {
+        Ok(recovery) => {
+            let output_slice = slice::from_raw_parts_mut(
+                recovery_shards_out,
```
asmie, line 35: Should we also check if that pointer is not null? `slice::from_raw_parts_mut` behavior is undefined if the pointer is null.

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+                recovery_shards_count as usize * shard_size as usize,
```
asmie, line 36: The same question for this multiplication - can it overflow?

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+/// - recovery_shards: number of recovery shards to generate
+/// - original_shards: input byte array containing flattened shards
+/// - shard_size: size of each shard in bytes
+/// - recovery_shards_out: buffer to store generated and flattened recovery shards
+/// Returns 0 on success, -1 on error
+#[no_mangle]
+pub unsafe extern "C" fn reed_solomon_encode(
+    original_shards_count: size_t,
+    recovery_shards_count: size_t,
+    shard_size: size_t,
```
asmie, line 17: The general question - maybe we could just assign this to some new variable and not use conversion to `usize` during each usage inside the function? 😸 
Or can we just skip `as usize` each time? The decode function doesn't have it.

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    original_shards_len: size_t,
+    original_shards_indexes: *const size_t,
+    recovery_shards: *const u8,
+    recovery_shards_len: size_t,
+    recovery_shards_indexes: *const size_t,
+    recovered_shards_out: *mut u8,
+    recovered_shards_indexes_out: *mut size_t,
+) -> c_int {
+    // Create original shard pairs
+    let orig_data = slice::from_raw_parts(original_shards, original_shards_len);
```
asmie, line 82: The same applies to all the pointers. Is there a guarantee that they won't be null?

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    let orig_indexes =
+        slice::from_raw_parts(original_shards_indexes, original_shards_len / shard_size);
```
asmie, line 84: What if `shard_size` is 0?

I wonder if we should have some sanity check at the beginning like:
```
    if shard_size == 0 || original_shards_len % shard_size != 0 || recovery_shards_len % shard_size != 0 {
        return -1;
    }
```

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+        recovery_inputs,
+    ) {
+        Ok(restored) => {
+            let shards_recovered_out =
+                slice::from_raw_parts_mut(recovered_shards_out, restored.len() * shard_size);
+
+            let shards_recovered_indexes_out =
+                slice::from_raw_parts_mut(recovered_shards_indexes_out, restored.len());
+
+            for (i, (&shard_index, shard_data)) in restored.iter().enumerate() {
```
asmie, line 112: I don't know how long can be `restored`, but if there are many entries, it might be worth considering a parallel iterator. (This can also be applied when copying slices in the `encode` function).

asmie (changes requested):
The Rust part.

Very good job! I like the copying slice into slice part, using `copy_from_slice` was the good choice.

The general remark is that it would be nice to add some tests to check if the interface is good. Just several simple cases showing that we cannot easily trigger undefined behaviour.

Example:
```
#[cfg(test)]
mod tests {
    use super::*;
    use std::ptr;

    #[test]
    fn test_decode_success() {
        // Mock data setup
        let original_shards_count = 2;
        let recovery_shards_count = 1;
        let shard_size = 4;

        let original_shards: [u8; 8] = [1, 2, 3, 4, 5, 6, 7, 8];
        let original_shards_indexes: [usize; 2] = [0, 1];

        let recovery_shards: [u8; 4] = [9, 10, 11, 12];
        let recovery_shards_indexes: [usize; 1] = [2];

        let mut recovered_shards_out: [u8; 4] = [0; 4];
        let mut recovered_shards_indexes_out: [usize; 1] = [0; 1];

        unsafe {
            let result = reed_solomon_decode(
                original_shards_count,
                recovery_shards_count,
                shard_size,
                original_shards.as_ptr(),
                original_shards.len(),
                original_shards_indexes.as_ptr(),
                recovery_shards.as_ptr(),
                recovery_shards.len(),
                recovery_shards_indexes.as_ptr(),
                recovered_shards_out.as_mut_ptr(),
                recovered_shards_indexes_out.as_mut_ptr(),
            );

            assert_eq!(result, 0); // Success
            assert_eq!(recovered_shards_out, [9, 10, 11, 12]);
            assert_eq!(recovered_shards_indexes_out, [2]);
        }
    }

    #[test]
    fn test_decode_invalid_shard_size() {
        let original_shards_count = 2;
        let recovery_shards_count = 1;
        let shard_size = 4;

        let original_shards: [u8; 9] = [1, 2, 3, 4, 5, 6, 7, 8, 9];
        let original_shards_indexes: [usize; 2] = [0, 1];

        let recovery_shards: [u8; 4] = [9, 10, 11, 12];
        let recovery_shards_indexes: [usize; 1] = [2];

        let mut recovered_shards_out: [u8; 4] = [0; 4];
        let mut recovered_shards_indexes_out: [usize; 1] = [0; 1];

        unsafe {
            let result = reed_solomon_decode(
                original_shards_count,
                recovery_shards_count,
                shard_size,
                original_shards.as_ptr(),
                original_shards.len(),
                original_shards_indexes.as_ptr(),
                recovery_shards.as_ptr(),
                recovery_shards.len(),
                recovery_shards_indexes.as_ptr(),
                recovered_shards_out.as_mut_ptr(),
                recovered_shards_indexes_out.as_mut_ptr(),
            );

            assert_eq!(result, -1); // Failure due to invalid shard size
        }
    }

    #[test]
    fn test_decode_mismatched_shard_counts() {
        let original_shards_count = 3; // Incorrect count
        let recovery_shards_count = 1;
        let shard_size = 4;

        let original_shards: [u8; 8] = [1, 2, 3, 4, 5, 6, 7, 8];
        let original_shards_indexes: [usize; 2] = [0, 1];

        let recovery_shards: [u8; 4] = [9, 10, 11, 12];
        let recovery_shards_indexes: [usize; 1] = [2];

        let mut recovered_shards_out: [u8; 4] = [0; 4];
        let mut recovered_shards_indexes_out: [usize; 1] = [0; 1];

        unsafe {
            let result = reed_solomon_decode(
                original_shards_count,
                recovery_shards_count,
                shard_size,
                original_shards.as_ptr(),
                original_shards.len(),
                original_shards_indexes.as_ptr(),
                recovery_shards.as_ptr(),
                recovery_shards.len(),
                recovery_shards_indexes.as_ptr(),
                recovered_shards_out.as_mut_ptr(),
                recovered_shards_indexes_out.as_mut_ptr(),
            );

            assert_eq!(result, -1); // Failure due to mismatched shard count
        }
    }

    #[test]
    fn test_decode_minimum_shards() {
        let original_shards_count = 1;
        let recovery_shards_count = 1;
        let shard_size = 4;

        let original_shards: [u8; 4] = [1, 2, 3, 4];
        let original_shards_indexes: [usize; 1] = [0];

        let recovery_shards: [u8; 4] = [5, 6, 7, 8];
        let recovery_shards_indexes: [usize; 1] = [1];

        let mut recovered_shards_out: [u8; 4] = [0; 4];
        let mut recovered_shards_indexes_out: [usize; 1] = [0; 1];

        unsafe {
            let result = reed_solomon_decode(
                original_shards_count,
                recovery_shards_count,
                shard_size,
                original_shards.as_ptr(),
                original_shards.len(),
                original_shards_indexes.as_ptr(),
                recovery_shards.as_ptr(),
                recovery_shards.len(),
                recovery_shards_indexes.as_ptr(),
                recovered_shards_out.as_mut_ptr(),
                recovered_shards_indexes_out.as_mut_ptr(),
            );

            assert_eq!(result, 0); // Success
            assert_eq!(recovered_shards_out, [5, 6, 7, 8]);
            assert_eq!(recovered_shards_indexes_out, [1]);
        }
    }

#[test]
fn test_decode_null_pointer() {
    let original_shards_count = 2;
    let recovery_shards_count = 1;
    let shard_size = 4;

    let original_shards: [u8; 8] = [1, 2, 3, 4, 5, 6, 7, 8];
    let original_shards_indexes: [usize; 2] = [0, 1];

    let recovery_shards: [u8; 4] = [9, 10, 11, 12];
    let recovery_shards_indexes: [usize; 1] = [2];

    let mut recovered_shards_out: [u8; 4] = [0; 4];
    let mut recovered_shards_indexes_out: [usize; 1] = [0; 1];

    unsafe {
        // Test null pointer for original_shards
        let result = reed_solomon_decode(
            original_shards_count,
            recovery_shards_count,
            shard_size,
            ptr::null(),
            original_shards.len(),
            original_shards_indexes.as_ptr(),
            recovery_shards.as_ptr(),
            recovery_shards.len(),
            recovery_shards_indexes.as_ptr(),
            recovered_shards_out.as_mut_ptr(),
            recovered_shards_indexes_out.as_mut_ptr(),
        );
        assert_eq!(result, -1); // Failure due to null pointer

        // Test null pointer for recovery_shards
        let result = reed_solomon_decode(
            original_shards_count,
            recovery_shards_count,
            shard_size,
            original_shards.as_ptr(),
            original_shards.len(),
            original_shards_indexes.as_ptr(),
            ptr::null(),
            recovery_shards.len(),
            recovery_shards_indexes.as_ptr(),
            recovered_shards_out.as_mut_ptr(),
            recovered_shards_indexes_out.as_mut_ptr(),
        );
        assert_eq!(result, -1); // Failure due to null pointer

        // Test null pointer for recovered_shards_out
        let result = reed_solomon_decode(
            original_shards_count,
            recovery_shards_count,
            shard_size,
            original_shards.as_ptr(),
            original_shards.len(),
            original_shards_indexes.as_ptr(),
            recovery_shards.as_ptr(),
            recovery_shards.len(),
            recovery_shards_indexes.as_ptr(),
            ptr::null_mut(),
            recovered_shards_indexes_out.as_mut_ptr(),
        );
        assert_eq!(result, -1); // Failure due to null pointer

        // Test null pointer for recovered_shards_indexes_out
        let result = reed_solomon_decode(
            original_shards_count,
            recovery_shards_count,
            shard_size,
            original_shards.as_ptr(),
            original_shards.len(),
            original_shards_indexes.as_ptr(),
            recovery_shards.as_ptr(),
            recovery_shards.len(),
            recovery_shards_indexes.as_ptr(),
            recovered_shards_out.as_mut_ptr(),
            ptr::null_mut(),
        );
        assert_eq!(result, -1); // Failure due to null pointer
    }
}
}
```

internal/erasurecoding/reedsolomon/reedsolomon.go
```diff
@@ -0,0 +1,247 @@
...
+		fmt.Println("Failed to load erasure coding library:", err)
+		os.Exit(1)
+	}
+
+	// Register the Rust FFI functions with Go using purego
+	purego.RegisterLibFunc(&reedSolomonEncode, lib, "reed_solomon_encode")
+	purego.RegisterLibFunc(&reedSolomonDecode, lib, "reed_solomon_decode")
+}
+
+func getErasurecodingLibaryPath() (string, error) {
```
pantrif, line 227: This function is quite similar to `getBandersnatchLibraryPath`. I suggest to move the method to a central/common location and passing the library name as a parameter.

internal/erasurecoding/reedsolomon/reedsolomon.go
```diff
@@ -0,0 +1,247 @@
...
+
+	if len(shards) != (r.originalShardsCount + r.recoveryShardsCount) {
+		return errors.New("not enough space for recovery shards")
+	}
+
+	shardSize := shardSize(shards)
+
+	flatOriginalShards := make([]byte, r.originalShardsCount*shardSize)
+	for i, s := range shards[:r.originalShardsCount] {
+		copy(flatOriginalShards[i*shardSize:], s)
```
bamzedev, line 99: Do we need to maybe check for size?
```
if len(s) != shardSize {
        return errors.New("inconsistent shard size")
    }

File: internal/erasurecoding/reedsolomon/reedsolomon.go [hunk 11]
```diff
@@ -0,0 +1,247 @@
...
+		os.Exit(1)
+	}
+
+	// Load the Rust shared library
+	lib, err := purego.Dlopen(libPath, purego.RTLD_NOW|purego.RTLD_GLOBAL)
+	if err != nil {
+		fmt.Println("Failed to load erasure coding library:", err)
+		os.Exit(1)
+	}
+
```
bamzedev, line 221: Don't we need to a defer func() to clean up?

erasurecoding/src/lib.rs
greywolve: I think we can do some extra checks on `original_shards_count` , which shouldn't be more than 65535. Also going to add `original_shards_len; so we can verify that the `shard_size` is correct. That should avoid an overflow.

erasurecoding/src/lib.rs
greywolve: Yeah I realized after we don't need it, forgot to update encode too. Good catch!

erasurecoding/src/lib.rs
greywolve: I meant to add a bunch of null checks, will do so. :)

erasurecoding/src/lib.rs
greywolve: We'll be dealing around 342 * 2 bytes most of the time, so parallel iterators probably aren't worth it yet.

internal/erasurecoding/reedsolomon/reedsolomon.go
greywolve: Good point. The only problem is the paths are a little different, so you'd also need to pass the path to the lib. I also think we'll probably need to have some sort of config in place later , or some more sophisticated way of finding the library by searching for a few common paths. I think this might be best to defer until then.

internal/erasurecoding/reedsolomon/reedsolomon.go
greywolve: I don't a need if we're exiting, and I don't see any cleanup examples in the purego docs. Do you have some example code?

internal/erasurecoding/reedsolomon/reedsolomon.go
greywolve: Good point, I guess for now we want correctness so I'll add this check. Later we might want to add some sort of verify function maybe that's called before hand and checks that the shards are correct in one go.

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,600 @@
...
+    recovery_shards_out_len: size_t,
+) -> c_int {
+    if let Some(sum) = original_shards_count.checked_add(recovery_shards_count) {
+        if sum > MAX_SHARDS {
+            return -1;
+        }
+    } else {
+        // Overflow.
+        return -1;
+    }
```
asmie, line 34: This could be a bit shorter

```suggestion
    match original_shards_count.checked_add(recovery_shards_count) {
        Some(sum) if sum <= MAX_SHARDS => {},
        _ => return -1,
    };
```

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,600 @@
...
+        original_shards,
+    ) {
+        Ok(recovery) => {
+            let output_slice =
+                slice::from_raw_parts_mut(recovery_shards_out, recovery_shards_count * shard_size);
+
+            for (i, shard) in recovery.iter().enumerate() {
+                let start = i * shard_size;
+                let end = start + shard_size;
+                output_slice[start..end].copy_from_slice(shard);
```
asmie, line 75: nit: `end` is basically not needed here. But if you like, keep it 😄 

```suggestion
                let start = i * shard_size;
                output_slice[start..start + shard_size].copy_from_slice(shard);
```

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,600 @@
...
+    recovered_shards_indexes_out: *mut size_t,
+) -> c_int {
+    if let Some(sum) = original_shards_count.checked_add(recovery_shards_count) {
+        if sum > MAX_SHARDS {
+            return -1;
+        }
+    } else {
+        // Overflow.
+        return -1;
+    }
```
asmie, line 123: It can be written shorter:

```suggestion
    match original_shards_count.checked_add(recovery_shards_count) {
        Some(total_shards) if total_shards <= MAX_SHARDS => (),
        _ => return -1,
    }
```

erasurecoding/src/lib.rs
```diff
@@ -0,0 +1,600 @@
...
+        Ok(restored) => {
+            let recovered_shards_out =
+                slice::from_raw_parts_mut(recovered_shards_out, restored.len() * shard_size);
+
+            let recovered_shards_indexes_out =
+                slice::from_raw_parts_mut(recovered_shards_indexes_out, restored.len());
+
+            for (i, (&shard_index, shard_data)) in restored.iter().enumerate() {
+                let start = i * shard_size;
+                let end = start + shard_size;
```
asmie, line 192: nit: like for the `encode` 😄

asmie (approved):
Nice job! 
Only some nitpicks left - feel free to use them or not 😄 

Just a random thought: if we can distinguish what went wrong maybe we could use more error consts than only `-1` - but if we just want to indicate error, we can even join those `if`s together, as all of the error conditions returns exactly the same value 👍 

Overall, LGTM, merge whenever Go side is approved 👍

internal/erasurecoding/reedsolomon/reedsolomon.go
bamzedev: I don't know, I just saw that there is a `purego.Dlclose` function. I was just curious.

internal/erasurecoding/reedsolomon/reedsolomon.go
```diff
@@ -0,0 +1,265 @@
...
+	if len(shards) != (r.originalShardsCount + r.recoveryShardsCount) {
+		return errors.New("not enough space for recovery shards")
+	}
+
+	shardSize := shardSize(shards)
+	if shardSize == 0 {
+		return errors.New("invalid shard size")
+	}
+
+	flatOriginalShards := make([]byte, r.originalShardsCount*shardSize)
```
bamzedev, line 103: We should probably check for overflow.

internal/erasurecoding/reedsolomon/reedsolomon.go
```diff
@@ -0,0 +1,265 @@
...
+// resulting shards contain the original shards and enough space for the
+// recovery shards. Shards can be passed directly to encode. Shards will have a
+// length of the original shards count + the recovery shards count. Data is not
+// copied, so the input data should not be modified after.
+func (r *Encoder) Chunk(data []byte) ([][]byte, error) {
+	// Need at least two bytes per chunk.
+	if len(data) < r.originalShardsCount*2 {
+		return nil, errors.New("data length too short")
+	}
+	shardSize := (len(data) + r.originalShardsCount - 1) / r.originalShardsCount
```
bamzedev, line 65: We should probably check for overflow.

greywolve:
Thanks for all the feedback. :) Addressed it all. Added some limitations for total shards and shard size, that should prevent any overflows too.

bamzedev (approved):

bamzedev (approved):

Issue #162, opened 2024-12-05
Core QUIC/TLS Infrastructure


bamzedev:
QUIC protocol setup
TLS 1.3 configuration
Ed25519 certificate handling
ALPN negotiation
Connection management

PR #192, opened 2024-12-19
Add quic transport layer

bamzedev:
https://github.com/zdave-parity/jam-np/blob/main/simple.md
Closes https://github.com/eigerco/strawberry/issues/162

This is just the connection part and opening a specific stream `CE 128: Block request`. There is a separate ticket for block request, this is just a demo implementation of it.
Added just the several tests for now, will add more later.
Check the  `cmd` folder on how to run the nodes.
How it works: 
```Initial Setup:

Generate Ed25519 key pair
Create TLS certificate using these keys
Setup QUIC transport with the cert and configurations
Register protocol handlers for different stream kinds (like BlockRequest)


Connection Establishment:


Node A starts listening on an address
Node B connects to Node A using QUIC
Both sides verify each other's certificates
A single QUIC connection is established between them


Making a Request:


Node B wants blocks from Node A
B creates a new stream within the existing QUIC connection
B writes stream kind (e.g., BlockRequest) as first byte
B writes the request message (hash, direction, etc.)


Handling the Request:


Node A accepts the stream
Reads first byte to determine stream kind
Finds appropriate handler for that stream kind
Handler reads request message
Handler processes request and writes response
Stream is closed after response is sent


Response Processing:


Node B reads response from the stream
Processes response or handles any errors
Stream is closed

### DIFF
--- a/pkg/network/transport/transport.go
+++ b/pkg/network/transport/transport.go
@@ -266,7 +266,7 @@ func (t *Transport) handleConnection(qConn quic.Connection) *Conn {
 		return nil
 	}
 
-	conn := t.manageConnection(peerKey, qConn)
+	conn := t.storeConnection(peerKey, qConn)
 
 	if err := t.config.Handler.OnConnection(conn); err != nil {
 		t.cleanup(peerKey)
@@ -279,8 +279,8 @@ func (t *Transport) handleConnection(qConn quic.Connection) *Conn {
 	return conn
 }
 
-// manageConnection handles connection storage and replacement
-func (t *Transport) manageConnection(peerKey ed25519.PublicKey, qConn quic.Connection) *Conn {
+// storeConnection handles connection storage and replacement
+func (t *Transport) storeConnection(peerKey ed25519.PublicKey, qConn quic.Connection) *Conn {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 

### CODE REVIEW

File: pkg/network/handlers/block_request.go [hunk 1]
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// Parse the message content into BlockRequestMessage
+	if len(msg.Content) < 37 { // 32 (hash) + 1 (direction) + 4 (maxBlocks)
+		return fmt.Errorf("message too short")
+	}
+
+	request := BlockRequestMessage{
+		Direction: msg.Content[32], // After hash
+		MaxBlocks: binary.LittleEndian.Uint32(msg.Content[33:37]),
```
pantrif, line 48: Shouldn't we be using the Jam codec for encoding?

pkg/network/cert/cert.go
```diff
@@ -12,32 +13,109 @@ import (
...
 }
 
+// Config contains the parameters needed for certificate generation.
+type Config struct {
+	// PublicKey is the Ed25519 public key to embed in the certificate
+	PublicKey ed25519.PublicKey
+	// PrivateKey is used to sign the certificate
+	PrivateKey ed25519.PrivateKey
+	// CertValidityPeriod defines how long the certificate remains valid
+	CertValidityPeriod time.Duration
```
greywolve, line 36: Curious, do we know yet how long these need to remain valid for?

pkg/network/cert/cert.go
```diff
@@ -12,32 +13,109 @@ import (
...
+	}
+	if now.After(cert.NotAfter) {
+		return fmt.Errorf("certificate has expired")
+	}
+	return nil
+}
+
+// ExtractPublicKey retrieves the Ed25519 public key from a certificate.
+// Returns an error if the certificate doesn't use an Ed25519 key.
+func (v *Validator) ExtractPublicKey(cert *x509.Certificate) (ed25519.PublicKey, error) {
```
greywolve, line 97: nit: feels like this might have been better as just a plain function. doesn't seem to fit being in a validator

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+	// ValidateKind checks if a stream kind byte is valid
+	ValidateKind(kindByte byte) error
+}
+
+// CertValidator performs TLS certificate validation and public key extraction
+type CertValidator interface {
+	// ValidateCertificate checks if a certificate meets required criteria
+	ValidateCertificate(cert *x509.Certificate) error
+	// ExtractPublicKey retrieves the Ed25519 public key from a certificate
+	ExtractPublicKey(cert *x509.Certificate) (ed25519.PublicKey, error)
```
greywolve, line 36: nit: not sure this method makes sense for a validator.

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+	ValidateConnection(tlsState tls.ConnectionState) error
+}
+
+// Config contains all configuration parameters for a Transport
+type Config struct {
+	PublicKey     ed25519.PublicKey  // Node's public key
+	PrivateKey    ed25519.PrivateKey // Node's private key
+	TLSCert       *tls.Certificate   // TLS certificate
+	ListenAddr    string             // Address to listen on
+	CertValidator CertValidator      // Certificate validator
```
greywolve, line 65: Couldn't `CertValidator` and `ConnectionHandler` rather be directly on `Transport` maybe?

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+
+// Transport manages QUIC connections and their lifecycles
+type Transport struct {
+	config       Config
+	listener     *quic.Listener
+	mu           sync.RWMutex
+	conns        map[string]*Conn // Active connections mapped by peer key
+	ctx          context.Context
+	cancel       context.CancelFunc
+	acceptLoopWg sync.WaitGroup // For clean shutdown of accept loop
```
greywolve, line 77: nit: given we only ever `.Add(1)` I wonder if a  done channel isn't simpler

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+				return fmt.Errorf("%w: %v", ErrInvalidCertificate, err)
+			}
+			if err := t.config.CertValidator.ValidateCertificate(c); err != nil {
+				return fmt.Errorf("%w: %v", ErrInvalidCertificate, err)
+			}
+			return nil
+		},
+	}
+
+	quicConn, err := quic.DialAddr(t.ctx, addr, tlsConf, &quic.Config{
```
greywolve, line 172: Shouldn't we also configure keep alive if this is going to be a long running connection?

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	t.mu.Lock()
+	if existingConn, exists := t.conns[string(peerKey)]; exists {
+		fmt.Println("Found existing connection, closing it")
+		// Close existing connection before replacing it
+		if err := existingConn.Close(); err != nil {
+			fmt.Printf("Failed to close existing connection: %v\n", err)
+		}
+		delete(t.conns, string(peerKey))
```
greywolve, line 268: Deleting here doesn't actually reclaim the memory in the map. Given this is long running code we just need to be sure this doesn't keep increasing over time.

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+	peerKey, err := t.config.CertValidator.ExtractPublicKey(tlsState.PeerCertificates[0])
+	if err != nil {
+		fmt.Printf("Failed to extract peer key: %v\n", err)
+		if cerr := qConn.CloseWithError(0, fmt.Sprintf("%s: %v", ErrInvalidCertificate.Error(), err)); cerr != nil {
+			fmt.Printf("Failed to close connection: %v\n", cerr)
+		}
+		return nil
+	}
+
+	t.mu.Lock()
```
greywolve, line 261: nit: Could this whole critical section perhaps be a method rather like `StoreConnection`?

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+		return nil
+	}
+
+	return conn
+}
+
+// Cleanup removes a connection from the map
+func (t *Transport) cleanup(peerKey ed25519.PublicKey) {
+	t.mu.Lock()
+	delete(t.conns, string(peerKey))
```
greywolve, line 292: This delete too, see earlier comment on reclaiming memory.

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// Stop gracefully shuts down the transport and all active connections.
+// Waits for the accept loop to finish before returning.
+func (t *Transport) Stop() error {
+	t.cancel()
+	if t.listener != nil {
+		t.listener.Close()
+	}
+	t.acceptLoopWg.Wait() // Wait for acceptLoop to finish
```
greywolve, line 147: How are we ensuring all the connections get closed aswell? (might have missed this)

greywolve (commented):
Still have some more to review but here's my initial review. :)

pkg/network/peer/peer.go
```diff
@@ -0,0 +1,56 @@
...
+	pubKey ed25519.PublicKey
+}
+
+// NewPeer creates a new peer instance from an established transport connection.
+// It wraps the connection with protocol-specific functionality using the provided manager.
+// Parameters:
+//   - conn: The underlying transport connection
+//   - pubKey: The peer's Ed25519 public key
+//   - protoManager: The protocol manager for handling streams
+func NewPeer(conn *transport.Conn, pubKey ed25519.PublicKey, protoManager *protocol.Manager) *Peer {
```
greywolve, line 30: nit: Do we need the pubKey as an extra param if we already have it on the conn?

pkg/network/peer/peer.go
```diff
@@ -0,0 +1,56 @@
...
+	return &Peer{
+		conn:      conn,
+		protoConn: protoManager.WrapConnection(conn),
```
greywolve, line 33: Why do we need to wrap the connection here again? Doesn't the transport `handleConnection` already do this when we connect? Couldn't we just get that  proto connection that's already in the manager?

greywolve: Also we don't seem to use the `conn` on the struct for anything, are we keeping it for sure sort of future use case?

pkg/network/cert/cert.go
bamzedev: Curious as well, no idea.

pkg/network/cert/cert.go
bamzedev: Agree.

pkg/network/transport/transport.go
bamzedev: Agree.

pkg/network/transport/transport.go
bamzedev: Intentionally wanted to separate concerns here. Cert (it's not a default cert, it's custom) and Conn handling are jam specific, so it would mix concerns here a bit, also are subject to change, while Transport just does one thing - takes care of quic.

pkg/network/transport/transport.go
bamzedev: I'm not sure, don't think all connections are going to be long lived.

pkg/network/transport/transport.go
bamzedev: Yeah, possibly could be an issue. We will probably profile at some point and see if this gets ugly. Then we can think of a solution. We can address now, but I am not sure how big of an issue this could be and the potential solutions are not nice either. Wdyt?

pkg/network/transport/transport.go
bamzedev: Yeah.

pkg/network/transport/transport.go
bamzedev: Yeah, makes sense.

pkg/network/peer/peer.go
bamzedev: No, you are right.

pkg/network/handlers/block_request.go
bamzedev: Yeah, probably. Block request is just for testing purpose, it's not a real implementation.

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+	go func() {
+		defer t.acceptLoopWg.Done()
+		t.acceptLoop()
+	}()
+	return nil
+}
+
+// Stop gracefully shuts down the transport and all active connections.
+// Waits for the accept loop to finish before returning.
+func (t *Transport) Stop() error {
```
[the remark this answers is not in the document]
bamzedev: Yeah, should close the connections. Thanks!

pkg/network/cert/cert.go
greywolve: Yeah manager would make this alot clearer :)

pkg/network/transport/transport.go
greywolve: I might be misunderstanding, why do those make sense on the Config being passed in?

pkg/network/peer/peer.go
bamzedev: This is a new separate protocol connection. The handleConnection is when other nodes connect to you. The `conn` is for future, yes.

pkg/network/transport/transport.go
greywolve: I guess we can cross this bridge when we get there

pkg/network/transport/transport.go
greywolve: Yeah probably just worth making a note that we should check on this later potentially.

pkg/network/transport/transport.go
bamzedev: The Config contains protocol-specific implementation. At a protocol level you decide what rules you follow and inject those into transport. I quite like dependency injection, easy way to keep things not tightly coupled. Are we speaking about different things?

greywolve: Ah so this is just a different way of doing options, ok that makes sense. :) No problem with the injecting part, just not used the config struct approach for interfaces, but that works too.

greywolve (dismissed):
LGTM aside from some nits.

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+
+	quicConn, err := quic.DialAddr(t.ctx, addr, tlsConf, &quic.Config{
+		EnableDatagrams: true,
+		MaxIdleTimeout:  MaxIdleTimeout,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrDialFailed, err)
+	}
+
+	conn := t.handleConnection(quicConn)
```
greywolve, line 180: I think this might be validating the cert twice? The `VerifyPeerCertificate` callback already covers the validation part in `handleConnection` i think.

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+			go t.handleConnection(conn)
+		}
+	}
+}
+
+// handleConnection processes a new QUIC connection
+func (t *Transport) handleConnection(qConn quic.Connection) *Conn {
+	tlsState := qConn.ConnectionState().TLS
+
+	if err := t.verifyPeerCert(tlsState.PeerCertificates); err != nil {
```
greywolve, line 237: I don't think we need this at all. I think the listener will use the `tlsConfig` 's verify callback to handle this.

pkg/network/transport/transport.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+
+			go t.handleConnection(conn)
+		}
+	}
+}
+
+// handleConnection processes a new QUIC connection
+func (t *Transport) handleConnection(qConn quic.Connection) *Conn {
+	tlsState := qConn.ConnectionState().TLS
```
greywolve, line 235: Might be good to note that this is an unstable API (has a warning on the struct), if we upgrade later it could break.

greywolve: Actually couldn't we just do the connection validation in the `VerifyPeerCertifcate` callback too? Since you only need the negotiated protocol.

greywolve: Actually seems you can also use the `VerifyConnection` callback on `tls.Config` to do this.

pantrif (approved):

greywolve (approved):

PR #196, opened 2024-12-30
Add variable sized data blob erasure coding.

### LINKED ISSUE #195: Variable sized data erasure coding

greywolve:
Closes #195


carlos-romano (approved):

internal/erasurecoding/erasurecoding_test.go
```diff
@@ -0,0 +1,180 @@
...
+			if err != nil {
+				t.Fatalf("Decode failed: %v", err)
+			}
+
+			require.Equal(t, data, dataOut)
+		})
+	}
+}
+
+func TestEncodeDecodeInitialTestVector(t *testing.T) {
```
bamzedev, line 96: Should we move this test to the integration folder to be consistent with the rest that tests test vectors? Also saw that we have the `readsolomon_test` also testing vectors.

greywolve: Do you feel like this is really an integration test if it's only testing one component?

We also only have one semi official test vector at the moment, maybe that can be moved to integration when the proper erasure coding vectors land?

greywolve: I could have probably just inlined that initial test vector

bamzedev: Imo if it is a test vector it should live with the other test vector cases. If you inline it you hardcode the test vector, right? It's called integration test, but that is just an incorrect naming. In reality it is a test vector folder.

greywolve: This isn't an official test vector technically, I'll remove the vectors folder and just put the json in the package. When the official vectors land then I'll add those tests in the integration tests folder.

greywolve: (basically I also want a test like this at the package level)

bamzedev (approved):

Issue #198, opened 2024-12-31
Assurance test vectors post state


bamzedev:
Verify Assurance test vectors post state. Currently we have a `TODO`.

PR #205, opened 2025-01-07
refactor: test assurance output

danielvladco:
https://github.com/eigerco/strawberry/issues/198


bamzedev (commented):
LGTM, just need to fix the build. Also can you remove `runAssuranceFunctions()` if we are not going to use it?

carlos-romano (approved):

carlos-romano (approved):

PR #216, opened 2025-01-13
Add block store

bamzedev:
This PR adds the functionality to store blocks. 
For `GetBlockSequence`, we have this info from the `CE 128: Block request`:
```
Ascending exclusive: The sequence of blocks in the response should start with a child of the given block, 
followed by a grandchild, and so on.
Descending inclusive: The sequence of blocks in the response should start with the given block, 
followed by its parent, grandparent, and so on.
```
Closes https://github.com/eigerco/strawberry/issues/206

### DIFF
--- a/internal/store/chain.go
+++ b/internal/store/chain.go
@@ -1,0 +1,203 @@
+package store
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"sync/atomic"
+
+	"github.com/eigerco/strawberry/internal/block"
+	"github.com/eigerco/strawberry/internal/crypto"
+	"github.com/eigerco/strawberry/pkg/db"
+	"github.com/eigerco/strawberry/pkg/db/pebble"
+)
+
+var (
+	ErrBlockNotFound = errors.New("block not found")
+	ErrChainClosed   = errors.New("chain store is closed")
+)
+
+const (
+	prefixHeader byte = iota + 1
+	prefixBlock
+)
+
+// Chain manages blockchain storage using a key-value store
+type Chain struct {
+	db     db.KVStore
+	closed atomic.Bool
+}
+
+// NewChain creates a new chain store using KVStore
+func NewChain(db db.KVStore) *Chain {
+	return &Chain{db: db}
+}
+
+// PutBlock stores a block and its header atomically
+func (c *Chain) PutBlock(b block.Block) error {
+	if c.closed.Load() {
+		return ErrChainClosed
+	}
+
+	// Create new batch for atomic operations
+	batch := c.db.NewBatch()
+	defer batch.Close()
+
+	headerHash, err := b.Header.Hash()
+	if err != nil {
+		return fmt.Errorf("hash header: %w", err)
+	}
+
+	// TODO: We should probably store the header here and refactor the Header file storage (AncestorStore)
+	// Store full block
+	blockBytes, err := b.Bytes()
+	if err != nil {
+		return fmt.Errorf("marshal block: %w", err)
+	}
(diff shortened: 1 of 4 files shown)


internal/store/chain.go
```diff
@@ -0,0 +1,203 @@
...
+
+// Close closes the chain store
+func (c *Chain) Close() error {
+	if !c.closed.CompareAndSwap(false, true) {
+		return nil
+	}
+	return c.db.Close()
+}
+
+// PrefixToString converts a prefix byte to a string
```
greywolve, line 185: This seems unused, is it going to be used later?

internal/store/chain.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/eigerco/strawberry/pkg/db/pebble"
+)
+
+var (
+	ErrBlockNotFound = errors.New("block not found")
+	ErrChainClosed   = errors.New("chain store is closed")
+)
+
+const (
+	prefixHeader byte = iota + 1
```
greywolve, line 21: prefixHeader isn't used yet, is this going to be used later? (just checking)

internal/store/chain.go
```diff
@@ -0,0 +1,203 @@
...
+	// Create new batch for atomic operations
+	batch := c.db.NewBatch()
+	defer batch.Close()
+
+	headerHash, err := b.Header.Hash()
+	if err != nil {
+		return fmt.Errorf("hash header: %w", err)
+	}
+
+	// TODO: We should probably store the header here and refactor the Header file storage (AncestorStore)
```
carlos-mateo, line 51: Maybe we could start adding issue ID when we add a TODO. This way we don't forget to add an issue that addresses the TODO. Even if it's a placeholder and it's never done, could be useful.

carlos-romano, line 51: Maybe we could start adding issue ID when we add a TODO. This way we don't forget to add an issue that addresses the TODO. Even if it's a placeholder and it's never done, could be useful.

internal/store/chain.go
bamzedev: Most likely. Logs, debug, anything.

internal/store/chain.go
bamzedev: We should. We need to refactor the Header to be using this.

internal/store/chain.go
greywolve: Can this ever be unknown? shouldn't that at least log some kind of error?

internal/store/chain.go
```diff
@@ -0,0 +1,203 @@
...
+	var blocks []block.Block
+	currentHash := startHash
+
+	for uint32(len(blocks)) < maxBlocks {
+		if ascending {
+			// For ascending (exclusive), skip first block
+			if currentHash != startHash {
+				blocks = append(blocks, currentBlock)
+			}
+			// Find children and take the first one
```
greywolve, line 147: How will we deal with forks? Maybe we could add a TODO here?

bamzedev: Initially thought of putting a `BestChild` mock func here, but I don't know how it will work, maybe at this point there will be no forks here, I don't know.

greywolve (approved):

internal/store/chain.go
bamzedev: Yes it can, you can call it with any value. It's just a utility function, It shouldn't log any error imo.

internal/store/chain.go
bamzedev: Good idea, although we don't create such detailed tickets in general. It might put people off writing a `TODO` since it would require to create a ticket.

carlos-romano (approved):

PR #224, opened 2025-01-16
Add header storage

bamzedev:
Closes https://github.com/eigerco/strawberry/issues/218.


internal/block/header.go
```diff
@@ -60,94 +56,3 @@ func HeaderFromBytes(data []byte) (Header, error) {
 	}
 	return header, nil
 }
-
-// AncestorStoreSingleton the in memory store for headers that need to be kept for 24 hours
-// TODO: Add 24 hours TTL
```
carlos-romano, line 65: Maybe we could create a ticket about this so we don't forget?

carlos-romano (approved):

internal/block/header.go
bamzedev: When we get to the DA layer, then we will have to deal with the TTL for all the things we store.

greywolve (approved):
LGTM, nice and clean

PR #227, opened 2025-01-16
refactor: accumulate upgrade to version 0.5.4 of GP

### LINKED ISSUE #209: accumulation changes 0.5.4
https://github.com/gavofyork/graypaper/compare/v0.5.3...v0.5.4#diff-3cefbb5fee9ce14d157e1da7c491ec078d4a61e4ba7d85a0a2d22655bee31853L223

danielvladco:
closes #209


internal/statetransition/state_transition.go
```diff
@@ -2423,6 +2424,14 @@ func (a *Accumulator) SequentialDelta(
 	return uint32(maxReports), newCtx, transfers, hashPairs
 }
 
+func getMapKeys[K comparable, V any](m map[K]V) map[K]struct{} {
```
greywolve, line 2427: Looks like you're turning a map into a set, maybe `toSet` would this a bit clearer? 

Are you planning on using this helper more in the future? (noticed it's only used once so far)

greywolve (approved):
Just a minor nit. :)

internal/statetransition/state_transition.go
danielvladco: we will probably need something like this in the future but for now it probably doesn't make sense to have a generic

greywolve (dismissed):

pantrif (approved):

pantrif (approved):

Issue #211, opened 2025-01-09
pvm invocation changes 0.5.4


bamzedev:
https://github.com/gavofyork/graypaper/compare/v0.5.3...v0.5.4#diff-3cefbb5fee9ce14d157e1da7c491ec078d4a61e4ba7d85a0a2d22655bee31853L223

Update Appendix B to match v0.5.4:

- [x] Update invocations and host calls
- [x] Remove Quit host function
- [x] Implement Eject host function
- [x] Implement Query host function
- [ ] Implement Yield host function

PR #231, opened 2025-01-20
Add eject host function

pantrif:
part of this https://github.com/eigerco/strawberry/issues/211

### DIFF
--- a/internal/polkavm/host_call/accumulate_functions.go
+++ b/internal/polkavm/host_call/accumulate_functions.go
@@ -1,6 +1,8 @@
 package host_call
 
 import (
+	"bytes"
+
 	"github.com/eigerco/strawberry/internal/block"
 	"github.com/eigerco/strawberry/internal/common"
 	"github.com/eigerco/strawberry/internal/crypto"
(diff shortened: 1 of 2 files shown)


internal/polkavm/host_call/accumulate_functions.go
```diff
@@ -253,9 +256,63 @@ func Eject(gas Gas, regs Registers, mem Memory, ctxPair AccumulateContextPair, t
...
+		return gas, withCode(regs, WHO), mem, ctxPair, nil
+	}
+
+	// if d ∈ K((x_u)_d)
+	serviceAccount, ok := ctxPair.RegularCtx.AccumulationState.ServiceState[block.ServiceId(d)]
+	if !ok {
+		return gas, withCode(regs, WHO), mem, ctxPair, nil
+	}
+
+	encodedXs, err := jam.Marshal(struct {
```
danielvladco, line 279: Why not just use the method like this? is it because of the `length=32` tag?
```suggestion
	encodedXs, err := jam.Marshal(ctxPair.RegularCtx.ServiceId)
```

danielvladco (approved):

internal/polkavm/host_call/accumulate_functions.go
pantrif: Yes the number of octets/bytes should be 32 based on GP. So I had to use the length tag

Issue #230, opened 2025-01-17
Validators grid structure


bamzedev:
Primarily for the purpose of block announcements, the previous, current, and next validator sets are conceptually arranged in a grid structure. Two validators are considered neighbours in the grid if:

They are validators in the same epoch and either have the same row (index / W) or the same column (index % W). W here is floor(sqrt(V)), where V is the number of validators.
They are validators in different epochs but have the same index.

PR #232, opened 2025-01-21
Add validator grid mapper for neighbor discovery

bamzedev:
Closes https://github.com/eigerco/strawberry/issues/230

```
Primarily for the purpose of block announcements, the previous, current, 
and next validator sets are conceptually arranged in a grid structure. 
Two validators are considered neighbours in the grid if:

They are validators in the same epoch and either have the same row (index / W) 
or the same column (index % W). W here is floor(sqrt(V)), where V is the number of validators.
They are validators in different epochs but have the same index.
```

### DIFF
--- a/internal/validator/grid.go
+++ b/internal/validator/grid.go
@@ -200,8 +200,6 @@ func getCurrentEpochNeighborIndices(validatorIndex uint16) []uint16 {
 	return neighbors
 }
 
-
-
 // areGridNeighbors determines if two validators within the same epoch are neighbors
 // in the grid structure by checking if they share the same row or column.
 // The grid width is calculated as floor(sqrt(total_validators)).
--- a/internal/validator/grid.go
+++ b/internal/validator/grid.go
@@ -45,9 +45,20 @@ func (m GridMapper) GetAllEpochsNeighborValidators(index uint16) ([]*crypto.Vali
 		return nil, fmt.Errorf("failed to get current epoch neighbor validators: %w", err)
 	}
 
-	// Add the two validators (from the previous epoch and the next epoch) with the same index
+	// Initialize with capacity for same epoch neighbors plus potentially two more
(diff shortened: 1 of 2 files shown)


greywolve (approved):
LGTM, only comment would be a randomized test would be nice too, but we can always add that later.

internal/validator/grid.go
```diff
@@ -0,0 +1,197 @@
...
+// neighbors within the same epoch based on the grid structure. This includes all
+// validators that share either:
+// - The same row (index / gridWidth)
+// - The same column (index % gridWidth)
+// The returned slice excludes the input validatorIndex itself.
+func getCurrentEpochNeighborIndices(validatorIndex uint16) []uint16 {
+	neighbors := make([]uint16, 0)
+
+	// Loop through all validators and check if they are neighbors
+	for i := uint16(0); i < common.NumberOfValidators; i++ {
```
pantrif, line 170: you can reduce iterations by directly compute row and column neighbours:
```
	// same row neighbours
	rowStart := (validatorIndex / gridWidth) * gridWidth
	for i := rowStart; i < rowStart+gridWidth; i++ {
		if i != validatorIndex {
			neighbors = append(neighbors, i)
		}
	}

	// Same column neighbors
	for i := validatorIndex % gridWidth; i < common.NumberOfValidators; i += gridWidth {
		if i != validatorIndex {
			neighbors = append(neighbors, i)
		}
	}
```

internal/validator/grid.go
```diff
@@ -0,0 +1,228 @@
...
+	for i := validatorIndex % gridWidth; i < common.NumberOfValidators; i += gridWidth {
+		if i != validatorIndex {
+			neighbors = append(neighbors, i)
+		}
+	}
+
+	return neighbors
+}
+
+func min(a, b uint16) uint16 {
```
pantrif, line 203: You can use the buildin `min` function

pantrif (approved):

PR #234, opened 2025-01-21
feat: upgrate pvm opcodes to v0.5.4

danielvladco:


internal/polkavm/instruction_codes.go
```diff
@@ -0,0 +1,170 @@
...
+
+package polkavm
+
+const (
+	// A.5.1. Instructions without Arguments
+	Trap        Opcode = 0 // trap
+	Fallthrough Opcode = 1 // fallthrough
+
+	// A.5.2. Instructions with Arguments of One Immediate.
+	Ecalli Opcode = 10 // ecalli
```
pantrif, line 11: Nit: Consider refactor this into groups of constants to leverage iota
Something like this:
```
const (
	// A.5.2. Instructions with Arguments of One Immediate.
	Ecalli Opcode = 10 + iota // 10
)

const (
	// A.5.3. Instructions with Arguments of One Register and One Extended Width Immediate.
	LoadImm64 Opcode = 20 + iota // 20
)

const (
	// A.5.4. Instructions with Arguments of Two Immediates.
	StoreImmU8  Opcode = 30 + iota // 30
	StoreImmU16
	StoreImmU32
	StoreImmU64
)
```

pantrif (approved):

internal/polkavm/instruction_codes.go
danielvladco: I like the idea with the groups of constants however I would like to define the opcodes specifically as they are defined in the graypaper, the fact that they are se incrementally is "just a coincidence" as I see it

PR #240, opened 2025-01-23
Review reports and implement exporting

carlos-romano:


internal/polkavm/common.go
```diff
@@ -298,3 +303,36 @@ type RefineContextPair struct {
 	IntegratedPVMMap map[uint64]IntegratedPVM //m
 	Segments         []Segment                //e
 }
+
+// ComputePagedProofs P(s) → [E(J₆(s,i), L₆(s,i))₍l₎ | i ∈ ℕ₍⌈|s|/64⌉₎] (14.10 v0.5.4)
+func ComputePagedProofs(r *RefineContextPair) ([]Segment, error) {
```
pantrif, line 308: the correct parameter here should be `[]Segment`

pantrif (approved):

PR #248, opened 2025-01-30
feat: add buffered decoder, bit sequence and decode with length

danielvladco:


greywolve (dismissed):

pkg/serialization/codec/jam/decode.go
```diff
@@ -50,6 +87,13 @@ func (br *byteReader) unmarshal(value reflect.Value) error {
 		return br.decodeFixedWidth(value, l)
 	case []byte:
 		return br.decodeBytes(value)
+	case []bool:
```
pantrif, line 90: I think we have to differentiate boolean slice from a bit sequence. I suggest creating another type for handling this case, maybe something like this: 
`type BitSequence []bool`

pkg/serialization/codec/jam/decode_test.go
```diff
@@ -0,0 +1,25 @@
+package jam
+
+import (
+	"bytes"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestDecodeBits(t *testing.T) {
+	input := []byte{17, 25, 0, 1, 2}
```
pantrif, line 10: can we add edge cases? single bit case or bit sequences not multiple of 8 (eg 10 bits)

pkg/serialization/codec/jam/decode.go
```diff
@@ -395,7 +443,27 @@ func (br *byteReader) decodeBytes(dstv reflect.Value) error {
...
+	for i := range *v {
+		mod := i % 8
+		if mod == 0 {
+			b, err = br.ReadOctet() // take as many bytes as needed to fill the bit sequence
+			if err != nil {
+				return err
+			}
+		}
+		pow2 := byte(1 << mod)   // powers of 2
+		(*v)[i] = b&pow2 == pow2 // identify the bit
```
pantrif, line 461: Is this ensuring least significant bit extraction?

pantrif (commented):
Shouldn't we also include the encoding part?

danielvladco:
> Shouldn't we also include the encoding part?
I didn't add it because we don't need it for now, and there are items with higher priority, plus I don't know if the bit sequence encoding is even used anywhere. I think at this point we should add things as we need them.

pkg/serialization/codec/jam/decode.go
danielvladco: the least significant bit will be `b&1<<0` when i % 8 == 0

pkg/serialization/codec/jam/decode.go
```diff
@@ -50,6 +89,13 @@ func (br *byteReader) unmarshal(value reflect.Value) error {
 		return br.decodeFixedWidth(value, l)
 	case []byte:
 		return br.decodeBytes(value)
+	case BitSequence:
```
pantrif, line 92: We handle aliases in the `handleReflectTypes` we should follow same convention here
```
	case reflect.Slice:
		if value.Type() == reflect.TypeOf(ed25519.PublicKey{}) {
			return br.decodeEd25519PublicKey(value)
		} else if value.Type() == reflect.TypeOf(BitSequence{}) {
			return br.decodeBit(value)
		}
		return br.decodeSlice(value)

danielvladco: Should this also have the byte slice in this case?

pantrif: yes you are right

pantrif (approved):

Issue #156, opened 2024-12-05
Work packages and Work Reports


bamzedev:

bamzedev:
Starting from 14.3.1 Exporting

pantrif:
We are missing the following components:
**Networking**
We need to implement QUIC communication for work-package submission (CE 133), sharing (CE 134), and work-report distribution (CE 135) to ensure that validators and guarantors can efficiently exchange data.

**Storage**
We need to persist work-package metadata, imported/exported segments, and justifications, ensuring that short-lived data (e.g auditable work-package) is retained until block finality, while long-lived data (e.g., exported segments and Paged-Proofs) is stored for at least 28 days.
Additionally, we might need to store work reports and refinement outputs, availability metadata, segment-root mappings, etc.

PR #249, opened 2025-01-30
Work results computation

pantrif:
Basic implementation (no networking or database) of work results computation section 14.4

part of this https://github.com/eigerco/strawberry/issues/156

### DIFF
--- a/internal/authorization/authorization.go
+++ b/internal/authorization/authorization.go
@@ -14,10 +14,6 @@ const (
 	isAuthorizedCost = 10
 )
 
-type Auth interface {
-	InvokePVM(workPackage work.Package, coreIndex uint16) ([]byte, error)
-}
-
 type EmptyContext struct{}
 
 type Authorization struct {
--- a/internal/authorization/authorization.go
+++ b/internal/authorization/authorization.go
@@ -14,6 +14,10 @@ const (
 	isAuthorizedCost = 10
 )
 
+type Auth interface {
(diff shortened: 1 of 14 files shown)


internal/authorization/authorization.go
```diff
@@ -14,6 +14,10 @@ const (
 	isAuthorizedCost = 10
 )
 
+type Auth interface {
```
greywolve, line 17: Couldn't we just have one eg `PVMInvoker` interface that's implemented by both auth and refine? And then this interface would just live in the results package.

greywolve: (also given this mostly seems in place for mocking)

internal/polkavm/host_call/refine_functions_test.go
```diff
@@ -341,7 +342,7 @@ func TestPeek(t *testing.T) {
 		IntegratedPVMMap: map[uint64]polkavm.IntegratedPVM{
 			n: u,
 		},
-		Segments: []polkavm.Segment{},
+		Segments: []work.Segment{},
```
carlos-romano, line 345: better

internal/authorization/authorization.go
pantrif: yes you are right! I moved them

carlos-romano (approved):

greywolve (approved):

greywolve (approved):

greywolve (approved):

PR #254, opened 2025-02-06
Updates to host calls v0.6.0

pantrif:
- Replaces OOB codes with panic on invalid memory access
- Offset for data-reading hostcalls

part of this https://github.com/eigerco/strawberry/issues/250

### DIFF
--- a/internal/polkavm/host_call/refine_functions.go
+++ b/internal/polkavm/host_call/refine_functions.go
@@ -43,31 +43,23 @@ func HistoricalLookup(
 		return gas, regs, mem, RefineContextPair{}, ErrAccountNotFound
 	}
 
-	ho := regs[A1]
-	bo := regs[A2]
-	bz := regs[A3]
+	// let [h, o] = ω8..+2
+	addressToRead, addressToWrite := regs[A1], regs[A2]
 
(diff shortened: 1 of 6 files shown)


internal/polkavm/host_call/refine_functions.go
```diff
@@ -402,9 +394,9 @@ func Invoke(
 	resultInstr, resultGas, resultRegs, resultMem, hostCall, invokeErr := interpreter.Invoke(program, pvm.InstructionCounter, invokeGas, invokeRegs, pvm.Ram)
 
 	if bb, err := jam.Marshal([14]uint64(append([]uint64{uint64(resultGas)}, resultRegs[:]...))); err != nil {
-		return gas, withCode(regs, OOB), mem, ctxPair, nil // (OOB, ω8, μ, m)
+		return gas, regs, mem, ctxPair, ErrPanicf(err.Error()) // (panic, ω8, μ, m)
```
danielvladco, line 397: ☇ 😄

pantrif: I prefer text for this, because the icon is not clearly visible as comment

PR #262, opened 2025-02-12
Add guarantee manager - no network/consensus

carlos-romano:


internal/work/results/guarantee.go
```diff
@@ -0,0 +1,164 @@
...
+
+	signature := ed25519.Sign(validatorPrivateKey, reportHash[:])
+
+	guarantee := &block.Guarantee{
+		WorkReport: *workReport,
+		Timeslot:   jamtime.CurrentTimeslot(),
+		Credentials: []block.CredentialSignature{
+			{
+				ValidatorIndex: validatorIndex,
+				Signature:      ([crypto.Ed25519SignatureSize]byte)(signature[:crypto.Ed25519SignatureSize]),
```
greywolve, line 128: Why do we need to limit the signature size here, wouldn't the signature be a fixed size anyway and what happens if it's bigger and gets cut off?

greywolve: Also we already seem to have `[crypto.Ed25519SignatureSize]byte` defined here: https://github.com/eigerco/strawberry/blob/main/internal/crypto/keys.go#L9

internal/block/guarantee.go
```diff
@@ -157,14 +157,22 @@ func NewErrorWorkResult(serviceId ServiceId, serviceHashCode, payloadHash crypto
 }
 
 func (w *WorkReport) Hash() (crypto.Hash, error) {
+	jsonData, err := w.Encode()
```
pantrif, line 160: This variable name is misleading, can you rename it to something else? (eg `encodedData`)

internal/work/results/guarantee.go
```diff
@@ -0,0 +1,164 @@
...
+
+// GuaranteeManager handles the generation and validation of guarantees for work packages.
+// Section 15 of the graypaper 0.6.1
+type GuaranteeManager struct {
+	computation *Computation
+}
+
+func NewGuaranteeManager(computation *Computation) *GuaranteeManager {
+	if computation == nil {
+		panic("computation cannot be nil")
```
pantrif, line 22: i would avoid panic at this level, return an error instead

internal/work/results/guarantee.go
```diff
@@ -0,0 +1,164 @@
...
+	"fmt"
+	"github.com/eigerco/strawberry/internal/block"
+	"github.com/eigerco/strawberry/internal/crypto"
+	"github.com/eigerco/strawberry/internal/jamtime"
+	"github.com/eigerco/strawberry/internal/state"
+	"github.com/eigerco/strawberry/internal/work"
+)
+
+// GuaranteeManager handles the generation and validation of guarantees for work packages.
+// Section 15 of the graypaper 0.6.1
```
pantrif, line 15: the latest gp version is 0.6.2

internal/work/results/guarantee.go
```diff
@@ -0,0 +1,164 @@
...
+		panic("computation cannot be nil")
+	}
+	return &GuaranteeManager{
+		computation: computation,
+	}
+}
+
+// ProcessWorkPackageGuarantee generates and validates a guarantee for a work package.
+// This handles the local validator operations described in section 15 of the JAM graypaper.
+// TODO: Missing (should be done in higher layers):
```
pantrif, line 31: can you also add an issue number or reference to JAMNP protocol (CE ***) of the things we are missing?

internal/work/results/guarantee.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, errors.New("public keys cannot be nil")
+	}
+
+	// Validate work package authorization
+	if err := gp.validateWorkPackageAuthorization(wp, coreIndex, authPool); err != nil {
+		return nil, fmt.Errorf("authorization validation failed: %w", err)
+	}
+
+	// Generate and validate work report.
+	workReport, err := gp.computation.EvaluateWorkPackage(wp, coreIndex)
```
pantrif, line 56: can you also add equation numbers as comment where possible?

pantrif (approved):

greywolve (approved):

Issue #200, opened 2025-01-02
CE 128: Block request


bamzedev:
Implement CE 128: Block request

PR #268, opened 2025-02-18
Add block request

bamzedev:
This PR adds network block request. The `finalization` part is still a mock, basically if there are 5 children we consider it finalized. It's not currently used, but it made sense to include it as the `service` is being used already in the block request. `Leaves`  are descendants of the latest finalized block with no known children (this is from UP 0: Block announcement).
Closes https://github.com/eigerco/strawberry/issues/200

### DIFF
--- a/internal/chain/service.go
+++ b/internal/chain/service.go
@@ -93,11 +93,11 @@ func (bs *BlockService) initializeState() error {
 		return fmt.Errorf("failed to store genesis block: %w", err)
 	}
 	bs.Mu.Lock()
+	defer bs.Mu.Unlock()
 	bs.LatestFinalized = LatestFinalized{
 		Hash:          hash,
 		TimeSlotIndex: genesisHeader.TimeSlotIndex,
 	}
-	bs.Mu.Unlock()
 	return nil
 }
 
@@ -111,11 +111,11 @@ func (bs *BlockService) initializeState() error {
 // Note: May return nil even if finalization isn't possible (e.g., missing ancestors).
 // This is due to genesis block handling and is not considered an error.
 func (bs *BlockService) checkFinalization(hash crypto.Hash) error {
-	// Start from current header and walk back 5 generations
+	// Start from current header and walk back 6 generations
 	currentHash := hash
 	var ancestorChain []block.Header
 
-	// Walk back 5 generations
+	// Walk back 6`` generations
 	for i := 0; i < 6; i++ {
 		header, err := bs.Store.GetHeader(currentHash)
 		if err != nil {
--- a/pkg/network/handlers/block_request.go
+++ b/pkg/network/handlers/block_request.go
@@ -2,7 +2,6 @@ package handlers
 
 import (
 	"context"
-	"encoding/binary"
 	"fmt"
 
 	"github.com/eigerco/strawberry/internal/block"
@@ -58,7 +57,7 @@ func (h *BlockRequestHandler) HandleStream(ctx context.Context, stream quic.Stre
 	// Read the request message
 	msg, err := ReadMessageWithContext(ctx, stream)
 	if err != nil {
-		return fmt.Errorf("failed to read request message: %w", err)
+		return fmt.Errorf("read request message: %w", err)
 	}
 
 	// Validate minimum message length:
@@ -66,11 +65,15 @@ func (h *BlockRequestHandler) HandleStream(ctx context.Context, stream quic.Stre
 	if len(msg.Content) < 37 { // 32 (hash) + 1 (direction) + 4 (maxBlocks)
 		return fmt.Errorf("message too short")
 	}
+	var maxBlocks uint32
+	if err = jam.Unmarshal(msg.Content[33:37], &maxBlocks); err != nil {
+		return fmt.Errorf("unmarshal maxBlocks: %w", err)
+	}
 
 	// Parse fixed-size wire format
 	request := blockRequestMessage{
 		Direction: msg.Content[32], // After hash
-		MaxBlocks: binary.LittleEndian.Uint32(msg.Content[33:37]),
+		MaxBlocks: maxBlocks,
 	}
 	copy(request.Hash[:], msg.Content[:32])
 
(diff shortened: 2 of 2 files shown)


pkg/network/handlers/block_request.go
```diff
@@ -4,107 +4,151 @@ import (
...
+}
+
+// blockRequestMessage represents the wire format for block requests.
+// As per protocol spec:
+// - Header Hash: Starting point for block sequence
+// - Direction: 0 for ascending (exclusive), 1 for descending (inclusive)
+// - MaxBlocks: Maximum number of blocks to return
+type blockRequestMessage struct {
+	Hash      crypto.Hash
+	Direction byte // 0 for ascending, 1 for descending
```
pantrif, line 39: It would be clearer to use a boolean here (eg `IsDescending`)

pkg/network/handlers/block_request.go
```diff
@@ -4,107 +4,151 @@ import (
...
+func (r *BlockRequester) RequestBlocks(ctx context.Context, stream quic.Stream, headerHash [32]byte, ascending bool, maxBlocks uint32) ([]block.Block, error) {
 	direction := byte(0)
 	if !ascending {
 		direction = 1
 	}
 	content := make([]byte, 37)
 	copy(content[:32], headerHash[:])
 	content[32] = direction
-	binary.LittleEndian.PutUint32(content[33:], 10)
+	binary.LittleEndian.PutUint32(content[33:], maxBlocks)
```
pantrif, line 130: you can use the Jam codec for serialisation instead of the binary package

internal/chain/service.go
```diff
@@ -0,0 +1,238 @@
...
+// It handles block announcements according to UP 0 protocol specification,
+// maintaining the set of leaf blocks and tracking finalization status.
+type BlockService struct {
+	Mu              sync.RWMutex
+	KnownLeaves     map[crypto.Hash]jamtime.Timeslot // Maps leaf block hashes to their timeslots
+	LatestFinalized LatestFinalized                  // Tracks the most recently finalized block
+	Store           *store.Chain                     // Persistent block storage
+}
+
+// LatestFinalized represents the latest finalized block in the chain.
```
greywolve, line 31: Do we only need to store the latest finalized block? (curious) Wouldn't we want to have a store of every finalized block hash too?

internal/chain/service.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	if err := bs.Store.PutBlock(b); err != nil {
+		return fmt.Errorf("failed to store genesis block: %w", err)
+	}
+	bs.Mu.Lock()
+	bs.LatestFinalized = LatestFinalized{
+		Hash:          hash,
+		TimeSlotIndex: genesisHeader.TimeSlotIndex,
+	}
+	bs.Mu.Unlock()
```
greywolve, line 100: Might be safer to just defer the unlock here.

internal/chain/service.go
```diff
@@ -0,0 +1,238 @@
...
+// Returns nil if finalization check succeeds, error if any operations fail.
+// Note: May return nil even if finalization isn't possible (e.g., missing ancestors).
+// This is due to genesis block handling and is not considered an error.
+func (bs *BlockService) checkFinalization(hash crypto.Hash) error {
+	// Start from current header and walk back 5 generations
+	currentHash := hash
+	var ancestorChain []block.Header
+
+	// Walk back 5 generations
+	for i := 0; i < 6; i++ {
```
greywolve, line 119: Isn't this 6 generations?

internal/chain/service.go
```diff
@@ -0,0 +1,238 @@
...
+	bs.Mu.Lock()
+	defer bs.Mu.Unlock()
+	bs.KnownLeaves[hash] = slot
+}
+
+// RemoveLeaf removes a block from the set of known leaves.
+func (bs *BlockService) RemoveLeaf(hash crypto.Hash) {
+	bs.Mu.Lock()
+	defer bs.Mu.Unlock()
+	delete(bs.KnownLeaves, hash)
```
greywolve, line 206: Should probably note that the delete here won't actually clean up the memory. Not sure how much this will fill up over time but it could be the source of a leak later. Not saying we need to try address this now, just that it's good to note.

internal/chain/service.go
```diff
@@ -0,0 +1,238 @@
...
+// isDescendantOfFinalized checks if a block is a descendant of the latest finalized block
+// by walking back through its ancestors until we either:
+// - Find the latest finalized block (true)
+// - Find a different block at the same height as latest finalized (false)
+// - Can't find a parent (error)
+func (bs *BlockService) isDescendantOfFinalized(header *block.Header) (bool, error) {
+	bs.Mu.RLock()
+	finalizedSlot := bs.LatestFinalized.TimeSlotIndex
+	finalizedHash := bs.LatestFinalized.Hash
+	bs.Mu.RUnlock()
```
greywolve, line 218: Is there a reason not to just defer the unlock too? (curious)

greywolve (dismissed):
Minor nits, otherwise seems fine to me. :)

pkg/network/handlers/block_request.go
bamzedev: This is how it comes from the protocol. I can do a translation, but this is used in only one place, feel redundant to add a transformation.

internal/chain/service.go
bamzedev: Just one I assume, that's why is considered finalized. Otherwise we store everything regardless, but that whole logic is probably going to change when we get a working grandpa.

internal/chain/service.go
bamzedev: This function is used basically to see if we are even going to bother with this block. We lock just to set the latest finalized, then there is not need to block for the whole duration of the function, even if the latest changes, it is still correct (will still be a descendant).

pkg/network/handlers/block_request.go
bamzedev: A bit more verbose tho, but probably makes sense to use it.

internal/chain/service.go
bamzedev: We spoke about it in a previous PR. At some point we will probably profile.

pantrif (approved):

PR #270, opened 2025-02-20
Authorizations test vectors

carlos-romano:


tests/integration/authorizations_test.go
```diff
@@ -0,0 +1,186 @@
...
+			for i := range expectedPostState.CoreAuthorizersPool {
+				expected := expectedPostState.CoreAuthorizersPool[i]
+				actual := newAuthPools[i]
+
+				require.ElementsMatch(t, expected, actual,
+					"Mismatch in CoreAuthorizersPool[%d]", i)
+			}
+
+			for i := range expectedPostState.PendingAuthorizersQueues {
+				expected := preState.PendingAuthorizersQueues[i]
```
bamzedev, line 152: I'm confused. Are the pre and post PendingAuthorizersQueues not changing?

carlos-romano: Correct. Not part of the purpose of the test as they change at another step of the logic.

I could not check it, but I thought it was descriptive.

internal/crypto/hash.go
```diff
@@ -22,3 +25,17 @@ func KeccakData(data []byte) Hash {
...
+// StringToHex converts a hex string to a byte slice
+func StringToHex(s string) []byte {
+	// Remove 0x prefix if present
+	s = strings.TrimPrefix(s, "0x")
+
+	// Decode hex string
+	bytes, err := hex.DecodeString(s)
+	if err != nil {
+		log.Printf("Error decoding hex string '%s': %v", s, err)
+		panic(err)
```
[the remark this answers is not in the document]
carlos-romano: haha it was an utility function for testing purposes, moved it here because of the comments on the meeting. Let me adapt to prod code standards :)

tests/integration/authorizations_test.go
bamzedev: Even more confused then :D, why are they putting them in the vectors then?

internal/crypto/hash.go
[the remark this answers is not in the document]
bamzedev: For clarification I meant the `panic`. I'd assume we will be using this in prod.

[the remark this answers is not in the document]
carlos-romano: yes I understood thanks! :) updated

tests/integration/authorizations_test.go
carlos-romano: same they did on reports test vectors with authorizations. They didn't modify them (although it would have changed if we would ahve updated state) but it was there.

bamzedev (approved):

greywolve (approved):

PR #272, opened 2025-02-22
Update block sealing

greywolve:
- Fix block sealing. We should have included the VRF signature before sealing.
- Add a function to verify block signatures.
- Add community test vector integration tests. (These were somehow left out of the past commit)
- Misc renaming for clarity.

### DIFF
--- a/internal/state/block_seal.go
+++ b/internal/state/block_seal.go
@@ -333,6 +333,7 @@ func VerifyBlockSignatures(
 			)
 			if ok {
 				publicKey = keys.Bandersnatch
+				break
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


internal/state/block_seal.go
```diff
@@ -251,3 +273,108 @@ func encodeUnsealedHeader(header block.Header) ([]byte, error) {
...
+			}
+
+			ok, _ := bandersnatch.Verify(
+				keys.Bandersnatch,
+				sealContext,
+				unsealedHeader,
+				header.BlockSealSignature,
+			)
+			if ok {
+				publicKey = keys.Bandersnatch
```
bamzedev, line 335:
```suggestion
				publicKey = keys.Bandersnatch
				break
```

internal/state/block_seal_test.go
bamzedev, line 1: Should we add couple of cases for invalid vrf and block seal signatures?

bamzedev (approved):

PR #274, opened 2025-02-26
CE 133 Work package submission

pantrif:
Partial implementation of CE-133 (since the availability system is not in place we cannot fetch the import segments yet)

Part of this https://github.com/eigerco/strawberry/issues/156

### DIFF
--- a/pkg/network/handlers/work_package_submission.go
+++ b/pkg/network/handlers/work_package_submission.go
@@ -90,7 +90,7 @@ func (h *WorkPackageSubmissionHandler) HandleStream(ctx context.Context, stream 
 	}
 
 	if err = stream.Close(); err != nil {
-		fmt.Printf("failed to close stream: %d", err)
+		return fmt.Errorf("failed to close stream: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


pkg/network/handlers/work_package_submission.go
```diff
@@ -0,0 +1,129 @@
...
+		return fmt.Errorf("failed to read message 1: %w", err)
+	}
+
+	var coreIndex uint16
+	if err = jam.Unmarshal(msg1.Content[:2], &coreIndex); err != nil {
+		return fmt.Errorf("failed to unmarshal core index: %w", err)
+	}
+
+	var pkg work.Package
+	if err = jam.Unmarshal(msg1.Content[2:], &pkg); err != nil {
```
greywolve, line 60: Just a thought, shouldn't we be trying to validate the size of the message before we unmarshal?

pantrif: I added a simple bound check for the message content

greywolve (dismissed):

pkg/network/handlers/work_package_submission.go
```diff
@@ -0,0 +1,133 @@
...
+			if err != nil {
+				// retry or reject package
+				continue
+			}
+			// Process or store segment data
+		}
+	}
+
+	if err = stream.Close(); err != nil {
+		fmt.Printf("failed to close stream: %d", err)
```
bamzedev, line 93: Should we return the error?

bamzedev: Also why `%d` ?

pantrif: fixed

bamzedev (approved):

PR #275, opened 2025-02-28
feat: add accumulation tiny test vectors

danielvladco:


internal/statetransition/state_transition.go
```diff
@@ -2423,6 +2417,7 @@ func (a *Accumulator) SequentialDelta(
 		}
 	}
 
+	log.Println("maxReports", maxReports)
```
greywolve, line 2420: Is this meant to be kept, or was it just for debugging?

danielvladco: already fixed it :D

tests/integration/accumulate_test.go
```diff
@@ -0,0 +1,381 @@
...
+		ReadyQueue: mapSlice(s.AccumulationQueue[:], func(t1 []state.WorkReportWithUnAccumulatedDependencies) []ReadyQueueItem {
+			return mapSlice(t1, func(t2 state.WorkReportWithUnAccumulatedDependencies) ReadyQueueItem {
+				var ss, pp []string
+				for d := range t2.Dependencies {
+					ss = append(ss, bytes2hex(d[:]))
+				}
+				for _, p := range t2.WorkReport.RefinementContext.PrerequisiteWorkPackage {
+					pp = append(pp, bytes2hex(p[:]))
+				}
+				//segmentRootLookup := make(map[crypto.Hash]crypto.Hash)
```
bamzedev, line 292: What's going on here?

tests/integration/accumulate_test.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func mapAccumulateState(t *testing.T, s AccumulateState) *state.State {
+	return &state.State{
+		Services: mapAccumulateServices(t, s.Accounts),
+		PrivilegedServices: service.PrivilegedServices{
+			ManagerServiceId:        s.Privileges.Bless,
+			AssignServiceId:         s.Privileges.Assign,
+			DesignateServiceId:      s.Privileges.Designate,
+			AmountOfGasPerServiceId: map[block.ServiceId]uint64{}, // s.Privileges.AlwaysAcc TODO is always empty should be mapped?
```
bamzedev, line 175: Does AlwaysAcc map to AmountOfGasPerServiceId? Seems strange.

tests/integration/accumulate_test.go
```diff
@@ -0,0 +1,381 @@
...
+			newState.AccumulationQueue,
+				newState.AccumulationHistory,
+				newState.Services,
+				newState.PrivilegedServices,
+				newState.ValidatorState.QueuedValidators,
+				newState.PendingAuthorizersQueues,
+				_ = statetransition.CalculateWorkReportsAndAccumulate(header, preState, newState.TimeslotIndex, workReports)
+
+			if !assert.Equal(t, postState, newState) {
+				d1, err := os.Create(file.Name() + "-post-state.json")
```
bamzedev, line 145: Is this meant to stay like this or was it for debugging only?

tests/integration/accumulate_test.go
danielvladco: it's for debugging, I removed it

bamzedev (approved):

tests/integration/accumulate_test.go
danielvladco: I mean rename according to the GP that being something like `AlwaysAccumulateServices`

bamzedev: Ah, ok, I had to check the GP. I see why they are named like this in the json. I'm not sure, you probably have the most knowledge here, w/e you think is best.

danielvladco: I think renaming it now is outside the scope of this PR, but we could consider renaming some fields in the future if we have time for refactors :)

greywolve (approved):

PR #276, opened 2025-03-05
Add block announcement

bamzedev:
Implements UP 0: Block announcement from JAMNP. Tests will be added later.

### DIFF
--- a/cmd/strawberry/main.go
+++ b/cmd/strawberry/main.go
@@ -3,40 +3,152 @@ package main
 import (
 	"context"
 	"crypto/ed25519"
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/eigerco/strawberry/internal/crypto"
+	"github.com/eigerco/strawberry/internal/safrole"
+	"github.com/eigerco/strawberry/internal/validator"
+	"github.com/eigerco/strawberry/pkg/network/peer"
 	"log"
 	"net"
-
-	"github.com/eigerco/strawberry/pkg/network/peer"
+	"os"
+	"strconv"
 )
 
-// main starts a blockchain node.
-// go run main.go -addr localhost:9000
-func main() {
-	ctx := context.Background()
-	listenAddr := flag.String("addr", "", "Listen address")
-	flag.Parse()
+type FullValidatorInfo struct {
+	Index      uint   `json:"index"`
+	IP         string `json:"address"`
+	Port       int    `json:"port"`
+	Ed25519Pub string `json:"ed25519_public_key"`
+	Ed25519Prv string `json:"ed25519_private_key"`
+}
 
-	if *listenAddr == "" {
-		log.Fatal("listen address is required")
-	}
-	// Generate node keys
-	pub, priv, err := ed25519.GenerateKey(nil)
+func (f FullValidatorInfo) ToJson() ([]byte, error) {
+	pub, prv, err := ed25519.GenerateKey(nil)
 	if err != nil {
 		panic(err)
 	}
-	keys := peer.ValidatorKeys{
-		EdPrv: priv,
-		EdPub: pub,
+	prvStr := hex.EncodeToString(prv)
+	pubStr := hex.EncodeToString(pub)
+	f.Ed25519Prv = prvStr
+	f.Ed25519Pub = pubStr
+	jsonData, err := json.MarshalIndent(f, "", "	")
+	if err != nil {
+		return nil, err
 	}
 
-	address, err := net.ResolveUDPAddr("", *listenAddr)
+	return jsonData, nil
+}
+
+func loadFullValidatorInfos(filename string) ([]FullValidatorInfo, error) {
+	jsonData, err := os.ReadFile(filename)
+	if err != nil {
+		return nil, fmt.Errorf("error reading file: %w", err)
+	}
+
+	var validators []FullValidatorInfo
+	if err := json.Unmarshal(jsonData, &validators); err != nil {
+		return nil, fmt.Errorf("error unmarshaling JSON: %w", err)
+	}
+
+	return validators, nil
+}
+
+func encodeToBytes(addr *net.UDPAddr) ([]byte, error) {
+	if len(addr.IP) != net.IPv6len {
+		return nil, fmt.Errorf("not an IPv6 address")
+	}
+
+	result := make([]byte, 18)
+	copy(result[:16], addr.IP)
+	binary.LittleEndian.PutUint16(result[16:], uint16(addr.Port))
+	return result, nil
+}
+
+func (f FullValidatorInfo) ToMetadata() ([]byte, error) {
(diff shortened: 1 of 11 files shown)


cmd/strawberry/main.go
```diff
@@ -3,47 +3,158 @@ package main
...
 
-	if *listenAddr == "" {
+	if *index == "" {
 		log.Fatal("listen address is required")
 	}
-	// Generate node keys
-	pub, priv, err := ed25519.GenerateKey(nil)
+	vs, err := loadFullValidatorInfos("test_validators.json")
 	if err != nil {
 		panic(err)
```
danielvladco, line 101: use `log.Fatal` everywhere in main

cmd/strawberry/main.go
```diff
@@ -3,47 +3,158 @@ package main
...
-	keys := peer.ValidatorKeys{
-		EdPrv: priv,
-		EdPub: pub,
+	i, err := strconv.Atoi(*index)
+	if err != nil {
+		panic(err)
 	}
-
-	address, err := net.ResolveUDPAddr("", *listenAddr)
+	address := vs[i].IP
```
danielvladco, line 107: prevent index out of bounds error

cmd/strawberry/main.go
```diff
@@ -3,47 +3,158 @@ package main
...
+	index := flag.String("index", "", "Validator Index")
 	flag.Parse()
 
-	if *listenAddr == "" {
+	if *index == "" {
 		log.Fatal("listen address is required")
 	}
-	// Generate node keys
-	pub, priv, err := ed25519.GenerateKey(nil)
+	vs, err := loadFullValidatorInfos("test_validators.json")
```
danielvladco, line 99: we should also provide a flag for this parameter instead of hardcoding it

internal/validator/manager.go
```diff
@@ -0,0 +1,92 @@
...
+// validators are considered neighbors if they share the same row or column in the grid.
+func (vm *ValidatorManager) IsNeighbor(key ed25519.PublicKey) bool {
+	return vm.GridMapper.IsNeighbor(vm.Keys.EdPub, key, true)
+}
+
+// IsSlotLeader determines if this validator is the designated leader for the current time slot.
+// It compares the validator's Bandersnatch public key with the key assigned to the current slot.
+// The fallbackKeys parameter provides the mapping of time slots to designated leader keys.
+// TODO: Probably need to have a way to termine the slot leader without the fallbackKeys.
+func (vm *ValidatorManager) IsSlotLeader(fallbackKeys crypto.EpochKeys) bool {
```
greywolve, line 67: I think I can add a function for this with the block sealing code, then this could just take `TicketsOrKeys`. But this is fine for now.

pkg/network/peer/node.go
```diff
@@ -273,6 +294,58 @@ func (n *Node) SubmitWorkPackage(ctx context.Context, coreIndex uint16, pkg work
...
+	if err != nil {
+		return fmt.Errorf("failed to get announcement handler: %w", err)
+	}
+
+	// Type assert to get the BlockAnnouncementHandler
+	bah, ok := handler.(*handlers.BlockAnnouncementHandler)
+	if !ok {
+		return fmt.Errorf("invalid handler type for block announcements")
+	}
+	// Check if we already have an announcer for this peer in BlockAnnouncementHandler. This should never happen.
```
greywolve, line 316: If this should never happen shouldn't we panic or at least log something?

cmd/strawberry/main.go
```diff
@@ -3,47 +3,158 @@ package main
...
+	return jsonData, nil
+}
+
+func loadFullValidatorInfos(filename string) ([]FullValidatorInfo, error) {
+	jsonData, err := os.ReadFile(filename)
+	if err != nil {
+		return nil, fmt.Errorf("error reading file: %w", err)
+	}
+
+	var validators []FullValidatorInfo
```
greywolve, line 52: So the expected file format is a json array of `FullValidatorInfo` objects?

greywolve (dismissed):

pkg/network/handlers/block_announcement.go
```diff
@@ -0,0 +1,506 @@
...
+	requestor           BlockRequestor                   // Used to request blocks after announcements
+	peerKey             ed25519.PublicKey                // Ed25519 key of the connected peer
+}
+
+// NewBlockAnnouncer creates a new announcer for a given stream and peer.
+// It initializes the announcer in the SendingHandshake state and registers
+// it in the handler's Announcers map using the peer's Ed25519 key.
+func (bh *BlockAnnouncementHandler) NewBlockAnnouncer(bs *chain.BlockService, ctx context.Context, stream quic.Stream, peerKey ed25519.PublicKey) *BlockAnnouncer {
+	announcerCtx, cancel := context.WithCancel(ctx)
+	bh.Mu.Lock()
```
pantrif, line 86: I think you should use `mu` here (since you are writing to announcers below). Do we need both mutexes (`mu` & `Mu`)?

pkg/network/handlers/block_announcement.go
```diff
@@ -0,0 +1,506 @@
...
+		return fmt.Errorf("process new block: %w", err)
+	}
+
+	// Request the full block using the CE 128 protocol
+	h, err := header.Hash()
+	if err != nil {
+		return fmt.Errorf("hash header: %w", err)
+	}
+
+	ctx, cancel := context.WithTimeout(ba.ctx, 6*time.Second) // Arbitrary timeout
```
pantrif, line 381: I suggest making the timeout a const to make it easier to configure in the future

internal/validator/manager.go
greywolve: Actually this function can surely get everything from the validator state, there's no need to pass in fallback keys or something. Just `IsSlotLeader()` is good enough here I think. I'll add a function you can call later to check this which will just take the state and the validator's private key.

pkg/network/handlers/block_announcement.go
bamzedev: Should be `mu`. The `Mu` is from `BlockService`. Good catch.

cmd/strawberry/main.go
bamzedev: Yes. It's just for testing, probably shouldn't have updated the `main.go`.

pkg/network/peer/node.go
bamzedev: I added it just in case, but even if it happens, nothing goes wrong here, so no need to terminate.

internal/validator/manager.go
bamzedev: Yeah, leaving it as is until we have that function.

cmd/strawberry/main.go
bamzedev: It's really just a testing func at this point, we will have a running main func soon™

pantrif (approved):

greywolve (approved):

PR #284, opened 2025-03-13
Add function for determining the slot leader

greywolve:


bamzedev:
Perhaps capitalize commit msg

PR #287, opened 2025-03-18
Add SAFROLE fallback simulation test

greywolve:
- Add a SAFROLE fallback simulation test using a well defined genesis state, block and validator keys.
- Also fix incorrect package name for the state merklization package.


greywolve:
Closes #287

tests/simulation/simulation_test.go
```diff
@@ -0,0 +1,363 @@
...
+		// Update the current state and block.
+		currentState.TimeslotIndex = currentTimeslot
+		currentState.EntropyPool = newEntropyPool
+		currentState.ValidatorState = newValidatorState
+
+		currentBlock = newBlock
+	}
+}
+
+// Produces and seals an empty block. TODO, move this to a better package an
```
bamzedev, line 131: What do we need to make this happen?

greywolve: I need to get it a bit more functionally complete, ie adding extrinsics (and their hash), which will come in future iterations.

PR #293, opened 2025-04-04
fix: shard data and compute erasure root correctly

danielvladco:
Introduces `shardData` that returns all the shards and hash/segment root pairs required for justification. This data can be later stored in memory and given to the assurers


internal/work/results/compute.go
```diff
@@ -200,53 +200,69 @@ func (c *Computation) computeAvailabilitySpecifier(
...
+		ErasureRoot:               binary_tree.ComputeWellBalancedRoot(auditableHashSegmentRootPairs, crypto.HashData),
+		SegmentRoot:               binary_tree.ComputeConstantDepthRoot(segmentsToByteSlices(exportedSegments), crypto.HashData),
+		SegmentCount:              uint16(len(exportedSegments)),
+	}, nil
+}
 
+// shardData shards the auditable work-package and segments, computes the hash for each auditable work-package shard and
+// the merkle root for each bundle of segment shards and joins them together
+// implements parts of eq. 14.16
+func shardData(auditableBlob []byte, exportedSegments []work.Segment) ([][]byte, [][][]byte, [][]byte, error) {
```
greywolve, line 221: Maybe we can document the outputs here a bit more clearly.

internal/work/results/compute.go
```diff
@@ -354,7 +370,7 @@ func (c *Computation) EvaluateWorkPackage(
 // ComputePagedProofs P(s) → [E(J₆(s,i), L₆(s,i))₍l₎ | i ∈ ℕ₍⌈|s|/64⌉₎] (14.10 v0.5.4)
 func ComputePagedProofs(segments []work.Segment) ([]work.Segment, error) {
 	if len(segments) == 0 {
-		return nil, fmt.Errorf("no segments provided")
+		return nil, nil
```
greywolve, line 373: This shouldn't be an error?

internal/work/results/compute.go
```diff
@@ -200,53 +200,69 @@ func (c *Computation) computeAvailabilitySpecifier(
...
-	shards, err := erasurecoding.Encode(padded)
+	auditableShards, err := erasurecoding.Encode(padded)
 	if err != nil {
-		return block.WorkPackageSpecification{}, err
+		return nil, nil, nil, err
 	}
-	if len(shards) == 0 {
-		return block.WorkPackageSpecification{}, nil
+	if len(auditableShards) == 0 {
+		return nil, nil, nil, nil
```
greywolve, line 229: 0 auditableShards are ok?

danielvladco: no, idk why it's there :)

internal/work/results/compute.go
danielvladco: yes, because it is possible that after the execution we get no segments, this should not fail the process

PR #294, opened 2025-04-09
feat: shard distribution protocol CE 137

danielvladco:
Implements CE 137 shard distribution protocol


pkg/network/handlers/availability_assurer_test.go
```diff
@@ -0,0 +1,118 @@
...
+	})
+	require.NoError(t, err)
+	s, err := conn.OpenStream()
+	require.NoError(t, err)
+
+	caller(s, ctx)
+
+	<-done
+
+	s.Close()
```
bamzedev, line 117: Should check for errors.

pkg/network/handlers/availability_assurer_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.NoError(t, err)
+
+	tlsConfig := &tls.Config{
+		Certificates:       []tls.Certificate{*tlsCert},
+		ClientAuth:         tls.RequireAnyClientCert,
+		MinVersion:         tls.VersionTLS13,
+		InsecureSkipVerify: true,
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2) //
```
bamzedev, line 79: Will this be enough for CI?

internal/validator/service.go
```diff
@@ -0,0 +1,22 @@
+package validator
+
+import (
+	"context"
+
+	"github.com/eigerco/strawberry/internal/crypto"
+)
+
+type ValidatorService interface {
```
greywolve, line 9: This interface is essentially only used for mocks? (just checking)

pkg/network/handlers/availability_assurer.go
```diff
@@ -0,0 +1,149 @@
...
+	bundleShardMsg, err := ReadMessageWithContext(ctx, stream)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to read bundle shard message: %w", err)
+	}
+	segmentShardMsg, err := ReadMessageWithContext(ctx, stream)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to read segment shard message: %w", err)
+	}
+
+	if len(segmentShardMsg.Content)%12 != 0 {
```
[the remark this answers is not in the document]
greywolve: What is the 12 here again?

pkg/network/handlers/availability_assurer.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, nil, nil, fmt.Errorf("invalid segment shard length (%d): %w", len(segmentShardMsg.Content), err)
+	}
+	for i := 0; i < len(segmentShardMsg.Content); i += 12 {
+		segmentShard = append(segmentShard, segmentShardMsg.Content[i:i+12])
```
greywolve, line 138: See note about making this a named variable above.

internal/validator/mock_service.go
```diff
@@ -0,0 +1,21 @@
+package validator
```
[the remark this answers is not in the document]
danielvladco: we need this here because the mock will be used in other packages as well, and it will be used for other protocols

internal/validator/service.go
danielvladco: This is for the eventual logic for assurers, but it helps to have an interface for testing yes

internal/validator/mock_service.go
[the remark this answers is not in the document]
bamzedev: Okay. Not sure if it makes sense or not, but just in case you didn't noticed this [folder](https://github.com/eigerco/strawberry/tree/main/pkg/network/mocks).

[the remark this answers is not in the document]
danielvladco: I noticed it, it is in the network folder, if you want to have a general package for mocks then it should be in pkg or internal

pkg/network/handlers/availability_assurer.go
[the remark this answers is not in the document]
danielvladco: a segment is 4104 bytes, after erasure coding it will be split into 1023 chunks of 12 bytes (because according to eq. H.6 k=6 and 2k = 12)

internal/validator/mock_service.go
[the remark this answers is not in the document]
danielvladco: I don't think it's a good idea to move the mocks package to a different folder as part of this PR, we can do it in another one, but for now the validator service mock can stay here

pkg/network/network_test.go
```diff
@@ -421,3 +424,69 @@ func TestTwoNodesSubmitWorkPackage(t *testing.T) {
...
+
+	// Submit the work package from node1 to node2
+	bundleShard, segmentShard, justification, err := node1.ShardDistributionSend(ctx, node2Peer.Ed25519Key, erasureRoot, shardIndex)
+	require.NoError(t, err)
+	assert.Equal(t, expectedBundleShard, bundleShard)
+	assert.Equal(t, expectedSegmentShard, segmentShard)
+	assert.Equal(t, expectedJustification, justification)
+
+	// Wait for the submission to be processed
+	time.Sleep(500 * time.Millisecond)
```
bamzedev, line 489: Don't the above asserts, guarantee that it was processed ?

danielvladco: true, I will remove that

pkg/network/handlers/shard_distribution.go
```diff
@@ -0,0 +1,163 @@
...
+	for _, just := range justification {
+		switch len(just) {
+		case 32: // one hash
+			justificationBytes = append(justificationBytes, 0)
+			justificationBytes = append(justificationBytes, just...)
+		case 64: // two hashes in case of a leaf
+			justificationBytes = append(justificationBytes, 1)
+			justificationBytes = append(justificationBytes, just...)
+		default:
+			return nil, fmt.Errorf("unexpected justification path value (should be either one ore two hashes): %v", len(just))
```
bamzedev, line 97:
```suggestion
			return nil, fmt.Errorf("unexpected justification path value (should be either one or two hashes): %v", len(just))
```

pkg/network/handlers/shard_distribution.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return justification, nil
+}
+
+// ShardDistributionSender handles outgoing CE-137 calls
+type ShardDistributionSender struct{}
+
+// ShardDistribution implements the sender side of the CE 137 protocol for more details check ShardDistributionHandler
+func (s *ShardDistributionSender) ShardDistribution(stream quic.Stream, ctx context.Context, erasureRoot crypto.Hash, shardIndex uint16) (bundleShard []byte, segmentShard [][]byte, justification [][]byte, err error) {
```
bamzedev, line 128:
```suggestion
func (s *ShardDistributionSender) ShardDistribution(ctx context.Context, stream quic.Stream, erasureRoot crypto.Hash, shardIndex uint16) (bundleShard []byte, segmentShard [][]byte, justification [][]byte, err error) {
```

bamzedev (dismissed):

PR #295, opened 2025-04-10
Add SAFROLE ticket simulation

### LINKED ISSUE #285: Simulate state - SAFROLE - ticket submission

greywolve:
Also:
- Add functions to Create and Verify ticket proofs.
- Add method to get a RingProver from the SAFROLE state. Used to generate ring signatures for ticket proof submission.
- Rearrange SAFROLE state transition code given issues found during simulation. The accumulator needed to be reset before new tickets are added.

Closes #285

### DIFF
--- a/internal/safrole/state.go
+++ b/internal/safrole/state.go
@@ -70,7 +70,7 @@ func (state State) RingProver(privateKey crypto.BandersnatchPrivateKey) (*bander
 		return nil, fmt.Errorf("private key is not a ring member")
 	}
 
-	ringProver, err := bandersnatch.NewRingProver(privateKey, ring, uint(proverIdx))
+	return bandersnatch.NewRingProver(privateKey, ring, uint(proverIdx))
 	if err != nil {
 		return nil, err
 	}
--- a/internal/safrole/state.go
+++ b/internal/safrole/state.go
@@ -37,11 +37,7 @@ func (state State) RingVerifier() (*bandersnatch.RingVrfVerifier, error) {
 	for i, vd := range state.NextValidators {
 		ring[i] = vd.Bandersnatch
 	}
-	ringVerifier, err := bandersnatch.NewRingVerifier(ring)
-	if err != nil {
-		return nil, err
-	}
-	return ringVerifier, nil
+	return bandersnatch.NewRingVerifier(ring)
 }
 
 // Takes a private bandersnatch key and returns a new RingVrfProver for this state.
@@ -71,10 +67,6 @@ func (state State) RingProver(privateKey crypto.BandersnatchPrivateKey) (*bander
 	}
 
 	return bandersnatch.NewRingProver(privateKey, ring, uint(proverIdx))
-	if err != nil {
-		return nil, err
-	}
-	return ringProver, nil
 }
 
 // SelectFallbackKeys selects the fallback keys for the sealing key series.
--- a/internal/safrole/state.go
+++ b/internal/safrole/state.go
@@ -1,6 +1,8 @@
 package safrole
 
 import (
+	"fmt"
+
 	"github.com/eigerco/strawberry/internal/block"
 	"github.com/eigerco/strawberry/internal/common"
 	"github.com/eigerco/strawberry/internal/crypto"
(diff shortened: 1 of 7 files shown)


internal/state/ticket.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Creates a ticket proof as per equation 6.29 in the graypaper (v0.6.4)
+// The ticket extrinic is a sequence of ticket proofs.
+func CreateTicketProof(state *State, privateKey crypto.BandersnatchPrivateKey, attempt uint8) (block.TicketProof, error) {
+	if attempt > common.MaxTicketAttempts {
+		return block.TicketProof{}, errors.New("attempts exceeded")
+	}
+
+	ringProver, err := state.ValidatorState.SafroleState.RingProver(privateKey)
+	defer ringProver.Free()
```
bamzedev, line 19: If `err`, ringProver will be `nil`, so this will probably panic.

internal/state/ticket.go
```diff
@@ -0,0 +1,63 @@
...
+	return ticket, nil
+}
+
+// Verifies a ticket proof as per equation 6.29 in the graypaper (v0.6.4)
+// Returns the output hash / identifier of the ticket if the proof is valid, or
+// an error if it is not.
+func VerifyTicketProof(state *State, ticket block.TicketProof) (crypto.BandersnatchOutputHash, error) {
+	// TODO: if this is too expensive to construct each time we should pass it in.
+	ringVerifier, err := state.ValidatorState.SafroleState.RingVerifier()
+	defer ringVerifier.Free()
```
bamzedev, line 46: Same as above comment.

internal/state/ticket.go
greywolve: They both check if it's nil before trying to free. :) (See [here](https://github.com/eigerco/strawberry/blob/main/internal/crypto/bandersnatch/bandersnatch.go#L303))

bamzedev: Can you check with a test?

greywolve: Add a test for the ring prover/verifier?

bamzedev: Just to check somehow if this works fine with RingProver returning an error.

greywolve: Ah I need to also be checking if r == nil, fixing

greywolve: Good catch btw, should be sorted now. :)

internal/safrole/state.go
```diff
@@ -42,6 +44,39 @@ func (state State) RingVerifier() (*bandersnatch.RingVrfVerifier, error) {
...
+		if pk == publicKey {
+			proverIdx = i
+			found = true
+		}
+	}
+	if !found {
+		return nil, fmt.Errorf("private key is not a ring member")
+	}
+
+	ringProver, err := bandersnatch.NewRingProver(privateKey, ring, uint(proverIdx))
```
danielvladco, line 73: This can be written like this for short, but if you check for error you could also wrap the error
```suggestion
	return bandersnatch.NewRingProver(privateKey, ring, uint(proverIdx))
```

internal/state/ticket.go
```diff
@@ -0,0 +1,63 @@
...
+// The ticket extrinic is a sequence of ticket proofs.
+func CreateTicketProof(state *State, privateKey crypto.BandersnatchPrivateKey, attempt uint8) (block.TicketProof, error) {
+	if attempt > common.MaxTicketAttempts {
+		return block.TicketProof{}, errors.New("attempts exceeded")
+	}
+
+	ringProver, err := state.ValidatorState.SafroleState.RingProver(privateKey)
+	defer ringProver.Free()
+	if err != nil {
+		return block.TicketProof{}, err
```
danielvladco, line 21: wrapping errors

greywolve: You'd prefer wrapping the errors in both cases?

danielvladco: yes

greywolve: Even if the error coming in is pretty clear?

danielvladco: in go errors don't have a stack trace, this is why wrapping them helps with debugging to know exactly in which place the error happened even if the error is pretty clear

greywolve: my understanding is you generally wrap when you want to add additional context, or if the error is unclear. in this case it will be "unable to create RingVrfProver", do you still think wrapping is worth it in that case?

greywolve: i don't think wrapping here makes sense, but I have added your suggestions for a simpler return :)

bamzedev (approved):

Issue #166, opened 2024-12-05
Work Package & Report Protocols


bamzedev:
CE 133: Work-package submission
CE 134: Work-package sharing
CE 135: Work-report distribution
CE 136: Work-report request

pantrif:
Remaining todo tasks:
- CE 133: Fetch import segments from availability system
- CE 134: 
     - Make sure that all import segments have been retrieved and construct a valid wp hash to segment root mapping
     - Fill the fields of `WorkPackageSpecification` in the work report

PR #296, opened 2025-04-11
Work package sharing handler CE-134

pantrif:
part of this https://github.com/eigerco/strawberry/issues/166


pkg/network/handlers/work_package_sharing.go
```diff
@@ -67,40 +95,82 @@ func (h *WorkPackageSharer) shareWorkPackageWithOtherGuarantors(ctx context.Cont
...
+				return
+			}
+
+			log.Printf("Received work-report hash and signature from peer %v:\n- Hash: %x\n- Signature: %x",
+				g.ValidatorIndex, response.WorkReportHash, response.Signature)
 		}(g)
 	}
 
+	// start local refinement in parallel
+	wg.Add(1)
```
greywolve, line 131: Are we planning on doing more in parallel later, or more work after while this is running? As this stands I don't see how making this parallel is adding any speedup.

pantrif: Sharing the work package with guarantors and running refinement should happen in parallel. In the future we could probably parallelize the refinement (maybe per work item or something like that)

greywolve: Ah my bad I missed the wait group right at the top. Makes sense. 👍

pkg/network/handlers/work_report.go
```diff
@@ -0,0 +1,74 @@
...
+// ProduceWorkReport runs the refine invocation and returns the work report along with the wp hash
+func ProduceWorkReport(
+	ctx context.Context,
+	refine refine.RefinePVMInvoker,
+	serviceState service.ServiceState,
+	authOutput []byte,
+	coreIndex uint16,
+	bundle work.PackageBundle,
+	segmentRootLookup map[crypto.Hash]crypto.Hash,
+) (*block.WorkReport, *crypto.Hash, error) {
```
greywolve, line 23: Is this a pointer because the work report might be nil after producing?

Also given that the work report has a `Hash()` method, is there any advantage to returning both the work report and it's hash here, as opposed to just the work report?

pantrif: I return nil when there's an error happy to drop the pointer if you wish.
I removed the hash from the return values

greywolve: Up to you, but a non pointer is probably a bit cleaner. :)

greywolve (approved):
LGTM, but I think it probably needs a pass from one other person

pkg/network/handlers/work_package_sharing_handler.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/eigerco/strawberry/internal/service"
+	"github.com/eigerco/strawberry/internal/work"
+	"github.com/eigerco/strawberry/pkg/serialization/codec/jam"
+)
+
+// WorkPackageSharingHandler processes incoming CE-134 streams
+// This handler is used by a guarantor who receives a work-package bundle from another guarantor.
+type WorkPackageSharingHandler struct {
+	privateKey   ed25519.PrivateKey
+	Auth         authorization.AuthPVMInvoker
```
danielvladco, line 23: can these fields be private?

pkg/network/handlers/work_package_sharing.go
```diff
@@ -9,19 +9,36 @@ import (
...
 	"github.com/eigerco/strawberry/internal/work"
 	"github.com/eigerco/strawberry/pkg/network/peer"
 	"github.com/eigerco/strawberry/pkg/network/protocol"
 	"github.com/eigerco/strawberry/pkg/serialization/codec/jam"
 )
 
 type WorkPackageSharer struct {
-	guarantors []*peer.Peer
+	guarantors   []*peer.Peer
+	Auth         authorization.AuthPVMInvoker
```
danielvladco, line 25: Same here, why do we export these fields?

pantrif: there is no need, i have made them private

pkg/network/handlers/work_package_sharing_handler.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, item := range bundle.Package.WorkItems {
+		for _, imp := range item.ImportedSegments {
+			if _, exists := lookup[imp.Hash]; !exists {
+				log.Printf("Missing mapping for imported segment hash %x", imp.Hash)
+				return fmt.Errorf("missing segment-root mapping for hash: %x", imp.Hash)
+			}
+		}
+	}
+
+	log.Print("Segment-root mappings verified")
```
danielvladco, line 149: is this meant to be here?

pantrif: Yes I have put logs in various places for tracking the workflow.

danielvladco (dismissed):

danielvladco (approved):

greywolve (approved):

pkg/network/handlers/work_report.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/eigerco/strawberry/internal/block"
+	"github.com/eigerco/strawberry/internal/crypto"
+	"github.com/eigerco/strawberry/internal/refine"
+	"github.com/eigerco/strawberry/internal/service"
+	"github.com/eigerco/strawberry/internal/work"
+)
+
+// ProduceWorkReport runs the refine invocation and returns the work report
+func ProduceWorkReport(
+	ctx context.Context,
```
bamzedev, line 16: We are not using the context for now?

pkg/network/handlers/work_package_sharing_handler.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+}
+
+// HandleStream implements the guarantor side of the CE-134 protocol.
+// It reads two messages:
+//  1. [Core Index ++ Segments-Root Mappings]
+//  2. [Work-Package Bundle]
+//
+// [Work-Report Hash ++ Ed25519 Signature].
+func (h *WorkPackageSharingHandler) HandleStream(ctx context.Context, stream quic.Stream, peerKey ed25519.PublicKey) error {
```
bamzedev, line 49: Should we not check if we are even supposed to do this work? What if I am not responsible for this core?
If we are skipping this for simplicity, we should add a TODO.

pkg/network/network_test.go
```diff
@@ -421,3 +444,99 @@ func TestTwoNodesSubmitWorkPackage(t *testing.T) {
...
+		mockAuthorizationInvoker{},
+		mockRefineInvoker{},
+		serviceState,
+	)
+	sharer.SetGuarantors([]*peer.Peer{
+		node1.PeersSet.GetByAddress(node2Addr.String()),
+	})
+
+	err = sharer.ValidateAndShareWorkPackage(ctx, coreIndex, bundle)
+	require.NoError(t, err)
```
bamzedev, line 520: Not sure if this test does anything, since most of the errors are basically just logging. Is there anything useful we can test here? Can we check what is happening on the other side maybe?

pkg/network/handlers/work_report.go
```diff
@@ -0,0 +1,69 @@
+package handlers
```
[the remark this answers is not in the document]
pantrif: I am not sure yet where this should reside, but I moved it for now under work/results

pkg/network/handlers/work_report.go
pantrif: It's not used at the time, maybe in the future we might need to handle timeouts etc. I removed it

bamzedev (approved):

PR #302, opened 2025-04-16
SAFROLE state transition fixes

greywolve:
- Tickets should be verified using the updated entropy pool and not the entropy pool from the prior state. (and hence the prior epoch)
- Add a custom test vector to verify this edge case. The vector submits a ticket in the first slot of a new epoch. If the old entropy pool is used in this case it will error with bad ticket proof.


tests/integration/vectors/safrole/custom-publish-ticket-first-slot-new-epoch.json
```diff
@@ -0,0 +1,497 @@
+{
+    "input": {
+        "slot": 12,
```
bamzedev, line 3: nit: this is meaningless, right? To avoid any confusion, should we make this `600`?

greywolve: Partially true, we can jump anywhere in the next epoch, but if we jump more then one epoch ahead then we go back to the fallback key sequence, as per [eq. 6.24](https://graypaper.fluffylabs.dev/#/68eaa1f/0ea2020ea202?v=0.6.4)

greywolve: I'll make it 15. :)

bamzedev (approved):

PR #306, opened 2025-04-22
CE 135 Work report distribution

pantrif:
part of this https://github.com/eigerco/strawberry/issues/166


pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,510 @@
...
+// Share the work-package bundle with other guarantors and collect signed responses.
+//
+// Guarantor -> Guarantor
+//
+// --> Core Index ++ Segments-Root Mappings
+// --> Work-Package Bundle
+// --> FIN
+// <-- Work-Report Hash ++ Ed25519 Signature
+// <-- FIN
+func (h *WorkReportGuarantor) shareWorkPackage(
```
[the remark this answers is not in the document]
pantrif: This needs to be called from the handler, creating a method on the node won't work here, since it would introduce cyclic dependencies

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,510 @@
...
+	if g.ValidatorIndex != nil {
+		validatorIndex = *g.ValidatorIndex
+	}
+
+	log.Printf("Received work-report hash and signature from peer %v:\n- Hash: %x\n- Signature: %x",
+		validatorIndex, response.WorkReportHash, response.Signature)
+
+	guaranteeCh <- guaranteeResponse{
+		WorkReportHash: response.WorkReportHash,
+		ValidatorIndex: validatorIndex,
```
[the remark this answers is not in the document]
pantrif: I added a check to skip processing if the validator index is empty

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,514 @@
...
+
+// Collect local refinement result and responses from other guarantors (via channel).
+// Filter only matching hashes, add local signature, and construct the full Guarantee.
+// If enough signatures are collected (>= 2) after the timeout (maxWaitTime), proceed to distribute it (CE-135).
+func (h *WorkReportGuarantor) processWorkReports(
+	ctx context.Context,
+	localResultCh <-chan localReportResult,
+	guaranteeRespCh <-chan guaranteeResponse,
+) {
+	localResult, remoteResponses, err := h.collectReports(ctx, guaranteeRespCh, localResultCh)
```
bamzedev, line 268: Imo this should be before entering `processWorkReports`. Here we are not processing, we still dont have them, we are collecting.

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,514 @@
...
+	if err != nil {
+		log.Printf("local refinement failed: %v", err)
+
+		return
+	}
+	if localResult.err != nil {
```
bamzedev, line 274: Isnt this the same as above? The only error that could be returned from the `collectReports` is the localResult err, no?

bamzedev: Also this is probably wrong here as well, even if we fail locally, there are still 2/3 valid resuts, shluldn't we still proceed?

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,514 @@
...
+	ctx context.Context,
+	remoteCh <-chan guaranteeResponse,
+	localCh <-chan localReportResult,
+) (localReportResult, []guaranteeResponse, error) {
+	var remoteResults []guaranteeResponse
+	var localRes localReportResult
+
+	localReceived := false
+
+	ctx, cancel := context.WithTimeout(ctx, maxWaitTime)
```
bamzedev, line 339: This should just be for the third validator not for the whole work.

pkg/network/handlers/work_report_distribution_handler.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read guarantee: %w", err)
+	}
+
+	var guarantee block.Guarantee
+	if err := jam.Unmarshal(msg.Content, &guarantee); err != nil {
+		return fmt.Errorf("failed to unmarshal guarantee: %w", err)
+	}
+
+	log.Printf("Received work report guarantee with %d signatures", len(guarantee.Credentials))
```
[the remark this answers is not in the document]
pantrif: I think validators will be able to include it in a block at this point or if it is already included by other validators/block-authors they will need it to keep up with the chain. I will add a TODO in this line.

pkg/network/peer/peer.go
```diff
@@ -20,6 +21,7 @@ type BlockAnnouncer interface {
 // PeerSet maintains mappings between peer identifiers
 // (Ed25519 keys, network addresses, validator indices) and Peer objects.
 type PeerSet struct {
+	mu sync.RWMutex
```
bamzedev, line 24: This was intentionally left without mutex. Whoever works with this has the responsibility to lock. Even in this very commit you have 2 locks. If you are locking above in guarantor, why is this needed?

pantrif: I removed it from peerSet, i used the mutex we have in node to avoid data race in test

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,569 @@
...
+			continue
+		}
+		if p := h.peerSet.GetByEd25519Key(v.Ed25519); p != nil {
+			peers = append(peers, p)
+		}
+	}
+	return peers
+}
+
+// mergePeers returns the union of two peer slices ensuring no duplicate ValidatorIndex entries
```
bamzedev, line 550: This is wrong, it will produce duplicates. A peer in current will have index `i`, but will be index `y` in the next.

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,639 @@
...
+// - A general timeout of (maxWaitTimeForCollectingReports) is applied to the entire collection process
+// - All signatures are grouped by hash to identify the largest group with matching work-report hashes
+// - Local failures are tolerated as long as enough remote responses are collected
+// - If no matching quorum is found after timeout or all responses, the caller must handle the failure
+// - Context cancellation causes an immediate exit
+func (h *WorkReportGuarantor) collectSignedReports(
+	ctx context.Context,
+	remoteCh <-chan guaranteeResponse,
+	localCh <-chan localReportResult,
+) ([]signedWorkReport, error) {
```
bamzedev, line 370: If here we return `map[crypto.Hash][]signedWorkReport` instead of `[]signedWorkReport`, you can save some redundant slice -> map -> slice in the`processWorkReports`.  You are already keeping hash in a map, you could directly build and return a map[crypto.Hash][]signedWorkReport as you append each report.

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,639 @@
...
+}
+
+// collectSignedReports gathers signed work-report hashes from both local refinement and remote guarantors.
+//
+// - It waits for at least two successful responses (local or remote) that agree on the same hash
+// - Once two matching hashes are received, start a timer for a possible third response
+// - A general timeout of (maxWaitTimeForCollectingReports) is applied to the entire collection process
+// - All signatures are grouped by hash to identify the largest group with matching work-report hashes
+// - Local failures are tolerated as long as enough remote responses are collected
+// - If no matching quorum is found after timeout or all responses, the caller must handle the failure
```
bamzedev, line 364: This is not correct. It's possible to not reach quorum here, but there will be no error (3 different results). Wrong comment or code?

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,639 @@
...
+			continue
+		}
+		if p := h.peerSet.GetByEd25519Key(v.Ed25519); p != nil {
+			peers = append(peers, p)
+		}
+	}
+	return peers
+}
+
+// mergePeers returns the union of two peer slices ensuring no duplicate entries based on Ed25519 public keys
```
bamzedev, line 622: Comment is misleading, it does more than that.

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,639 @@
...
+func mergePeers(a, b []*peer.Peer) []*peer.Peer {
+	exists := make(map[string]struct{})
+	var merged []*peer.Peer
+
+	for _, p := range a {
```
bamzedev, line 627: This loop adds the first slice peers, even when they are not validators. So slice `a` everyone, slice `b` only validators.

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,639 @@
...
+			// Timer expired
+			return signedReports, nil
+		}
+
+		// if all responses arrived, return early
+		if totalResponses == maxSignaturesRequiredToGuarantee {
+			return signedReports, nil
+		}
+
+		// Check if quorum has been reached for any hash to start the timer
```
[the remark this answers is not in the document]
pantrif: done

pkg/network/handlers/work_report_guarantor.go
pantrif: I rephrased the comment

pkg/network/handlers/work_report_guarantor.go
pantrif: I have added more details + renamed the function

pkg/network/handlers/work_report_guarantor.go
```diff
@@ -0,0 +1,514 @@
...
+	if err != nil {
+		guaranteeCh <- guaranteeResponse{Err: err}
+		log.Printf("Failed to open stream to peer %v: %v", validatorIndex, err)
+		return
+	}
+
+	err = h.sendWorkPackage(ctx, stream, coreIndex, segments, bundle)
+	if err != nil {
+		guaranteeCh <- guaranteeResponse{Err: err}
+		log.Printf("Failed to share WP with peer %v: %v", validatorIndex, err)
```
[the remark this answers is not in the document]
bamzedev: `Failed to share` implies we failed during sending. `Failed to complete work package exchange with peer` is probably a bit more precise.

pkg/network/peer/common.go
```diff
@@ -0,0 +1,35 @@
...
+// MergeValidators returns a deduplicated set of validator peers from slices a and b,
+// including only peers that are validators (ValidatorIndex != nil)
+// Deduplication is performed based on Ed25519 public keys
+func MergeValidators(a, b []*Peer) []*Peer {
+	exists := make(map[string]struct{})
+	var merged []*Peer
+
+	// Include only valid peers from `a`
+	for _, p := range a {
+		if p.ValidatorIndex == nil {
```
bamzedev, line 12: Check for nil, so we don't panic here.

pkg/network/peer/common.go
```diff
@@ -0,0 +1,35 @@
...
+		key := string(p.Ed25519Key)
+		if _, seen := exists[key]; !seen {
+			exists[key] = struct{}{}
+			merged = append(merged, p)
+		}
+	}
+
+	// Include only valid, non-duplicate peers from `b`
+	for _, p := range b {
+		if p.ValidatorIndex == nil {
```
bamzedev, line 24: Check for nil, so we don't panic here.

pkg/network/peer/common.go
bamzedev, line 1: Do we need a separate file for a single func?

pkg/network/peer/common_test.go
```diff
@@ -0,0 +1,83 @@
...
+			a:        []*peer.Peer{newTestPeer(k1, &i1), newTestPeer(k2, &i2)},
+			b:        nil,
+			expected: []ed25519.PublicKey{k1, k2},
+		},
+		{
+			name:     "only validators from B",
+			a:        nil,
+			b:        []*peer.Peer{newTestPeer(k1, &i1), newTestPeer(k2, &i2)},
+			expected: []ed25519.PublicKey{k1, k2},
+		},
```
bamzedev, line 50:
```suggestion
		},
		{
			name: "inputs with nil peer entries",
			a:    []*peer.Peer{newTestPeer(k1, &i1), nil, newTestPeer(k2, &i2)},
			b:    []*peer.Peer{nil, newTestPeer(k3, &i3), newTestPeer(k1, &i1) /*duplicate*/},
			expected: []ed25519.PublicKey{
				k1, k2, k3,
			},
		},
		{
			name:     "all non-validators",
			a:        []*peer.Peer{newTestPeer(k1, nil), newTestPeer(k2, nil)},
			b:        []*peer.Peer{newTestPeer(k3, nil)},
			expected: nil, // or []ed25519.PublicKey{}
		},
		{
			name:     "duplicates within A, B is empty",
			a:        []*peer.Peer{newTestPeer(k1, &i1), newTestPeer(k2, &i2), newTestPeer(k1, &i1)}, // k1 is duplicated
			b:        nil,
			expected: []ed25519.PublicKey{k1, k2},
		},
		{
			name:     "duplicates within B, A is empty",
			a:        nil,
			b:        []*peer.Peer{newTestPeer(k1, &i1), newTestPeer(k2, &i2), newTestPeer(k1, &i1)}, // k1 is duplicated
			expected: []ed25519.PublicKey{k1, k2},
		},
```

pkg/network/peer/common.go
bamzedev: You should probably add this to `Peer`.
`func (p *Peer) IsValidator() bool {
	return p.ValidatorIndex != nil
}`

pkg/network/peer/common.go
pantrif: No I moved it to peer

pkg/network/peer/common.go
pantrif: Done

bamzedev (approved):

PR #307, opened 2025-04-23
Add multiple trie storage

bamzedev:
This PR contains some fixes for Patricia Merkle Trie:
- Misunderstanding of flags showing type of leaf
- Misunderstanding of how the hashes are trimmed
- Not storing the actual value of the leaf
Added a way to store mutliple tries. All nodes are stored separately and not linked to tries to avoid duplications.
Added a delete trie and node that prunes the unused nodes.
A little refactor of the store folder.

These fixes were needed in order to start https://github.com/eigerco/strawberry/issues/229

### DIFF
--- a/internal/store/chain.go
+++ b/internal/store/chain.go
@@ -65,30 +65,43 @@ func (c *Chain) FindHeader(fn func(header block.Header) bool) (block.Header, err
 	if err != nil {
 		return block.Header{}, fmt.Errorf("create iterator: %w", err)
 	}
-	defer func() {
-		if cerr := iter.Close(); cerr != nil && err == nil {
-			err = fmt.Errorf("iterator close error: %w", cerr)
-		}
-	}()
+
+	// We'll track any error we might return, to also handle close errors.
+	var returnErr error
+	var result block.Header
 
 	// Iterate through headers
 	for iter.Next() {
 		headerBytes, err := iter.Value()
 		if err != nil {
-			return block.Header{}, fmt.Errorf("get header value: %w", err)
+			returnErr = fmt.Errorf("get header value: %w", err)
+			break
 		}
 
 		header, err := block.HeaderFromBytes(headerBytes)
 		if err != nil {
-			return block.Header{}, fmt.Errorf("parse header from bytes: %w", err)
+			returnErr = fmt.Errorf("parse header from bytes: %w", err)
+			break
 		}
 
 		if fn(header) {
-			return header, nil
+			result = header
+			break
 		}
 	}
 
-	return block.Header{}, nil
+	// Close iterator explicitly
+	if cerr := iter.Close(); cerr != nil {
+		if returnErr == nil {
+			returnErr = fmt.Errorf("iterator close error: %w", cerr)
+		}
+	}
+
+	// If retErr is still nil, check if result was set
+	if returnErr != nil {
+		return block.Header{}, returnErr
+	}
+	return result, nil
 }
 
 // PutBlock stores a block and its header atomically
@@ -178,18 +191,19 @@ func (c *Chain) FindChildren(parentHash crypto.Hash) ([]block.Block, error) {
 	if err != nil {
 		return nil, fmt.Errorf("create iterator: %w", err)
 	}
-	defer iter.Close()
+
+	var resultErr error
 
 	// Iterate through blocks
 	for iter.Next() {
 		blockBytes, err := iter.Value()
 		if err != nil {
-			log.Println("read block value from iterator", err)
+			log.Println("read block value from iterator:", err)
 			continue
 		}
 		b, err := block.BlockFromBytes(blockBytes)
 		if err != nil {
-			log.Println("parse block from bytes", err)
+			log.Println("parse block from bytes:", err)
 			continue
 		}
 
@@ -198,7 +212,13 @@ func (c *Chain) FindChildren(parentHash crypto.Hash) ([]block.Block, error) {
 		}
 	}
 
-	return children, nil
+	// Close iterator manually and capture any error
+	if cerr := iter.Close(); cerr != nil {
+		resultErr = fmt.Errorf("iterator close error: %w", cerr)
+	}
+
+	// Return collected children and potential close error
+	return children, resultErr
 }
 
 // GetBlockSequence retrieves a sequence of blocks.
--- a/internal/store/trie.go
+++ b/internal/store/trie.go
@@ -234,14 +234,14 @@ func (t *Trie) deleteNode(hash crypto.Hash, forceDelete bool) error {
 		}
 
 		// Delete left child if not empty
-		if !isZeroHash(leftHash) {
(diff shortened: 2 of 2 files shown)


internal/store/chain.go
```diff
@@ -15,30 +14,20 @@ import (
...
-const (
-	prefixHeader byte = iota + 1
-	prefixBlock
 )
 
 // Chain manages blockchain storage using a key-value store
 type Chain struct {
-	db     db.KVStore
-	closed atomic.Bool
+	db.KVStore
```
greywolve, line 21: What's the advantage of embedding here? (embedding tends to have a lot more footguns to be aware of)

internal/store/chain.go
```diff
@@ -15,30 +14,20 @@ import (
...
-
-const (
-	prefixHeader byte = iota + 1
-	prefixBlock
 )
 
 // Chain manages blockchain storage using a key-value store
 type Chain struct {
-	db     db.KVStore
-	closed atomic.Bool
```
greywolve, line 29: Was this never really needed?

internal/store/chain.go
```diff
@@ -75,16 +60,16 @@ func (c *Chain) GetHeader(hash crypto.Hash) (block.Header, error) {
...
-
 	// Create iterator for header prefix
-	iter, err := c.db.NewIterator([]byte{prefixHeader}, []byte{prefixHeader + 1})
+	iter, err := c.NewIterator([]byte{prefixHeader}, []byte{prefixHeader + 1})
 	if err != nil {
 		return block.Header{}, fmt.Errorf("create iterator: %w", err)
 	}
-	defer iter.Close()
+	defer func() {
+		if cerr := iter.Close(); cerr != nil && err == nil {
```
greywolve, line 69: Is`err == nil` here even needed? Won't we return with an error from the line above if it's not nil?

internal/store/trie.go
```diff
@@ -0,0 +1,290 @@
...
+	"github.com/eigerco/strawberry/pkg/db/pebble"
+)
+
+const (
+	// ErrNotLeafNode is returned when a non-leaf node is used where a leaf is expected
+	ErrNotLeafNode = "not a leaf node"
+)
+
+type Trie struct {
+	db.KVStore
```
greywolve, line 20: Similar comment to further up, why embed?

internal/store/trie.go
```diff
@@ -0,0 +1,290 @@
...
+		closeErr := batch.Close()
+		if closeErr != nil {
+			return crypto.Hash{}, fmt.Errorf("merklize error: %v, failed to close batch: %v", err, closeErr)
+		}
+		return crypto.Hash{}, fmt.Errorf("merklize error: %v", err)
+	}
+
+	// Commit the batch
+	if err := batch.Commit(); err != nil {
+		closeErr := batch.Close()
```
greywolve, line 55: Same as my earlier comment, shouldn't these be deferred? But maybe pebble works differently and it's not critical to Close() ?

internal/store/chain.go
bamzedev: It's needed in order to not override the original error if occurred. If both errors have occurred this will return the original one, otherwise it will override it.

internal/store/chain.go
bamzedev: There was a close fn that is now gone and it was used there, but overall both shouldn't have need added.

internal/store/trie.go
```diff
@@ -0,0 +1,290 @@
...
+			return fmt.Errorf("failed to delete ref count for hash %x: %v", hash, err)
+		}
+		// Ref count already deleted, continue
+	}
+
+	return nil
+}
+
+// isZeroHash checks if a hash is all zeros
+func isZeroHash(hash crypto.Hash) bool {
```
greywolve, line 283: Given crypto.Hash is an array isn't it simpler to just use `x == crypto.Hash{}`, or predefine the zero hash as `crypto.Hash{}` .

internal/store/chain.go
greywolve: Ahh right for errors further down.

internal/store/trie.go
bamzedev: Left over from initial version where it was a loop checking bits, but ignoring the first 1.

greywolve (approved):

greywolve:
We really need some sort of test vectors to see if we get the right state serialization.

internal/store/chain.go
bamzedev: Tbh this was done quickly to solve the lint errors of code I didnt touch. I would should have probably done it like in the trie code, where I manually close on each error path and avoid such things.

internal/store/chain.go
```diff
@@ -108,48 +93,72 @@ func (c *Chain) FindHeader(fn func(header block.Header) bool) (block.Header, err
 
 // PutBlock stores a block and its header atomically
 func (c *Chain) PutBlock(b block.Block) error {
-	if c.closed.Load() {
-		return ErrChainClosed
-	}
 	// Create new batch for atomic operations
-	batch := c.db.NewBatch()
-	defer batch.Close()
```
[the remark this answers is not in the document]
bamzedev: Every error path closes it and this is the way to avoid complex defers.

internal/store/trie.go
bamzedev: Every error path closes it and this is the way to avoid complex defers.

internal/store/common.go
```diff
@@ -0,0 +1,40 @@
...
+const (
+	prefixHeader byte = iota + 1
+	prefixBlock
+	prefixTrieNode
+	prefixTrieNodeValue
+	prefixTrieNodeRefCount
+)
+
+// PrefixToString converts a prefix byte to a string
+func PrefixToString(p byte) string {
```
pantrif, line 17: this is currently unused, are you planning to use it during fetching or elsewhere?

internal/store/trie.go
```diff
@@ -0,0 +1,290 @@
...
+	} else if errors.Is(err, pebble.ErrNotFound) {
+		// First reference to this node
+		newCount = 1
+	} else {
+		// Unexpected error
+		return fmt.Errorf("failed to get ref count for hash %x: %v", hash, err)
+	}
+
+	countBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(countBytes, newCount)
```
pantrif, line 142: Is there a specific reason you're not using the JAM codec in this store?

internal/store/common.go
bamzedev: Mostly logging probably when/if :D

internal/store/trie.go
bamzedev: Yeah, this isn't related to anything outside of just being a simple counter, is not subject to `gav` changing stuff and is simpler.

pantrif (approved):

pantrif (approved):

PR #312, opened 2025-05-02
Add fetch state trie range

bamzedev:
CE129 is a protocol that handles State requests. It allows nodes to request a range of a block's posterior state between two keys with a size limit. The protocol returns:
A series of key/value pairs from the state trie
Boundary nodes to verify the returned data

The FetchStateTrieRange function implements:
Accepting the exact parameters specified in CE129 (start key, end key, max size)
Returning key/value pairs within the specified range
Providing boundary nodes covering paths from root to start key and to the last returned key
Enforcing size limits as required by the protocol

This is data part of https://github.com/eigerco/strawberry/issues/229. After this is merged, the last PR would be to request and send that data.

### DIFF
--- a/internal/store/trie.go
+++ b/internal/store/trie.go
@@ -1,6 +1,7 @@
 package store
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


internal/store/trie.go
```diff
@@ -278,3 +277,293 @@ func (t *Trie) deleteNode(hash crypto.Hash, forceDelete bool) error {
 
 	return nil
 }
+
+// FetchStateTrieRange retrieves a range of key-value pairs from the trie starting at startKey and ending at or before endKey.
+// It also returns boundary nodes covering the paths from root to the start key and to the last key in the response.
+// The response size is limited to maxSize bytes, unless the response contains only a single key/value pair.
+func (t *Trie) FetchStateTrieRange(rootHash crypto.Hash, startKey, endKey [31]byte, maxSize uint32) (keys [][31]byte, values [][]byte, boundaryNodes []trie.Node, err error) {
```
pantrif, line 284: Would it make sense to group key/value into a struct instead of returning two slices? 
```
type Pairs struct {
  Key   [31]byte
  Value []byte
}

File: internal/store/trie.go [hunk 2]
```diff
@@ -278,3 +277,293 @@ func (t *Trie) deleteNode(hash crypto.Hash, forceDelete bool) error {
...
+	// Check if startKey > endKey, return empty result
+	if bytesGreaterThan(startKey[:], endKey[:]) {
+		return [][31]byte{}, [][]byte{}, []trie.Node{}, nil
```
pantrif, line 287: you might need to validate maxSize as well (`>0`)

internal/store/trie.go
```diff
@@ -278,3 +277,293 @@ func (t *Trie) deleteNode(hash crypto.Hash, forceDelete bool) error {
...
+						lastKeyPath = firstKeyPath
+					}
+
+					// Break out of the loop - we've hit our size limit
+					break
+				}
+			}
+
+			// Early termination: if we've passed the end key, we can stop
+			if bytesGreaterThan(key31[:], endKey[:]) {
```
greywolve: nit: might be better to put this early termination at the top, seems easier to read that way (unless there's a specific reason it needs to be checked last)

internal/store/trie.go
bamzedev: Yes, added.
```// TrieRangeResult represents the result of a trie range query
type TrieRangeResult struct {
	Keys          [][31]byte  // The keys in the range
	Values        [][]byte    // The values corresponding to the keys
	BoundaryNodes []trie.Node // Boundary nodes covering the paths from root to the edges
}

File: internal/store/trie.go [hunk 4]
```diff
@@ -278,3 +277,293 @@ func (t *Trie) deleteNode(hash crypto.Hash, forceDelete bool) error {
...
+	return bytesCompare(a, b) > 0
+}
+
+// bytesLessThan returns true if a < b
+func bytesLessThan(a, b []byte) bool {
+	return bytesCompare(a, b) < 0
+}
+
+// bytesCompare compares two byte slices lexicographically
+func bytesCompare(a, b []byte) int {
```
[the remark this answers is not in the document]
bamzedev: Haven't thought about it. Using it now.

internal/store/trie.go
bamzedev: Yeah, should be at top, moved it now.

greywolve (dismissed):

pantrif (dismissed):

greywolve (approved):

pantrif (approved):

PR #314, opened 2025-05-06
fix activitiy statistics int types

greywolve:
Validator statistics fields should all be uint32.


greywolve:
See the asn types [here](https://github.com/davxy/jam-test-vectors/blob/polkajam-vectors/jam-types-asn/jam-types.asn#L257).

bamzedev (approved):

Issue #164, opened 2024-12-05
Block Distribution Protocols


bamzedev:
UP 0: Block announcement
CE 128: Block request
CE 129: State request

PR #319, opened 2025-05-13
Add state request

bamzedev:
Adds a state request network protocol CE129. This is closing https://github.com/eigerco/strawberry/issues/229 and https://github.com/eigerco/strawberry/issues/164.
```Key = [u8; 31] (First 31 bytes of key only)
Maximum Size = u32
Boundary Node = As returned by B/L, defined in the State Merklization appendix of the GP
Value = len++[u8]

Node -> Node

--> Header Hash ++ Key (Start) ++ Key (End) ++ Maximum Size
--> FIN
<-- [Boundary Node]
<-- [Key ++ Value]
<-- FIN
```
Added some additional tests for the trie.

### DIFF
--- a/pkg/network/handlers/state_request.go
+++ b/pkg/network/handlers/state_request.go
@@ -1,0 +1,237 @@
+package handlers
+
+import (
+	"bytes"
+	"context"
+	"crypto/ed25519"
+	"encoding/binary"
+	"fmt"
+	"github.com/eigerco/strawberry/internal/crypto"
+	"github.com/eigerco/strawberry/internal/merkle/trie"
+	"github.com/eigerco/strawberry/internal/store"
+	"github.com/eigerco/strawberry/pkg/serialization/codec/jam"
+	"github.com/quic-go/quic-go"
+)
+
+const (
+	hashSize         = crypto.HashSize
+	keySize          = 31
+	maxSizeSize      = 4
+	headerHashOffset = 0
+	keyStartOffset   = headerHashOffset + hashSize
+	keyEndOffset     = keyStartOffset + keySize
+	maxSizeOffset    = keyEndOffset + keySize
+)
+
+// StateRequestHandler processes incoming state request streams according to the CE 129 protocol.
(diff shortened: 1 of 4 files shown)


pkg/network/handlers/state_request.go
```diff
@@ -0,0 +1,237 @@
...
+		return fmt.Errorf("read request message: %w", err)
+	}
+	// Validate minimum message length:
+	// 32 bytes for HeaderHash, 31 bytes for KeyStart, 31 bytes for KeyEnd, and 4 bytes for MaxSize
+	if len(msg.Content) < crypto.HashSize+keySize+keySize+4 {
+		return fmt.Errorf("message too short")
+	}
+
+	// Decode the MaxSize as a little-endian uint32
+	maxSize := binary.LittleEndian.Uint32(msg.Content[maxSizeOffset : maxSizeOffset+maxSizeSize])
```
greywolve, line 72: Can't we just use the jam codec to decode this?

pkg/network/handlers/state_request.go
```diff
@@ -0,0 +1,237 @@
...
+func NewStateRequestHandler(trie *store.Trie) *StateRequestHandler {
+	return &StateRequestHandler{
+		trie: trie,
+	}
+}
+
+// stateRequestMessage represents the structure of an incoming state request.
+type stateRequestMessage struct {
+	HeaderHash crypto.Hash   // The hash of the block header whose state is being requested
+	KeyStart   [keySize]byte // The first key in the requested range (inclusive)
```
greywolve, line 43: nit: this might be clearer as a specific type, like `TrieKey` , but it's fine like this too for now.

pkg/network/handlers/state_request.go
bamzedev: It's the same value atm. It's a bit more consistent to use Jam codec, but it also return an err, the network it self works with LE, so somewhat decoupled and this saves us from `Gav` changes to the codec.

PR #320, opened 2025-05-13
Shard assignment

pantrif:
### Shard assignment

Erasure coded shards are assigned to validators as follows:

```math
i = (cR + v) \bmod V
```

Where:

- $v$ is the index of a validator.
- $i$ is the index of the shard assigned to the validator.
- $c$ is the index of the core which produced the work-report.
- $R$ is the recovery threshold: the minimum number of EC shards required to recover the original
  data. With 1023 validators, $R = 342$.
- $V$ is the number of validators.

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -282,7 +282,7 @@ func (n *Node) SubmitWorkPackage(ctx context.Context, coreIndex uint16, pkg work
 
 // ShardDistributionSend implements the sending side of the CE 137, it opens a connection to the provided peer
 // allowing the assurers request shards from the guarantor
-func (n *Node) ShardDistributionSend(ctx context.Context, peerKey ed25519.PublicKey, report block.WorkReport) (bundleShard []byte, segmentShard [][]byte, justification [][]byte, err error) {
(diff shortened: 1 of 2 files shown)


pkg/network/node/node.go
```diff
@@ -281,7 +282,7 @@ func (n *Node) SubmitWorkPackage(ctx context.Context, coreIndex uint16, pkg work
 
 // ShardDistributionSend implements the sending side of the CE 137, it opens a connection to the provided peer
 // allowing the assurers request shards from the guarantor
-func (n *Node) ShardDistributionSend(ctx context.Context, peerKey ed25519.PublicKey, erasureRoot crypto.Hash, shardIndex uint16) (bundleShard []byte, segmentShard [][]byte, justification [][]byte, err error) {
+func (n *Node) ShardDistributionSend(ctx context.Context, peerKey ed25519.PublicKey, report block.WorkReport) (bundleShard []byte, segmentShard [][]byte, justification [][]byte, err error) {
```
bamzedev, line 285: Not sure this makes sense tbh. It's way cleaner, yes, but passing all that data (the whole report) just for a `uint` and a `hash` seems too wasteful.

bamzedev (approved):

PR #322, opened 2025-05-15
fix: validator data pointer

danielvladco:


greywolve (approved):

internal/state/block_seal.go
```diff
@@ -325,10 +325,6 @@ func VerifyBlockSignatures(
 		// currentValidators and seeing which validator's bandersnatch public
 		// key is able to verify the seal signature.
 		for _, keys := range currentValidators {
-			if keys == nil {
```
bamzedev, line 328: Here and is some places bellow such checks are now removed. I don't know the implications, but I'd assume they were needed. Why not add a func to `ValidatorKey` `IsEmpty` or something and just replace it?

greywolve (changes requested):
As per @bamzedev 's review seems there's some places with nil checks that we need to make sure doesn't break anything

internal/statetransition/state_transition.go
```diff
@@ -2171,7 +2171,7 @@ func CalculateIntermediateCoreFromAssurances(validators safrole.ValidatorsData,
 // validateAssurancesSignature (127) ∀a ∈ EA ∶ as ∈ Eκ′[av ]e ⟨XA ⌢ H(E(Hp, af ))⟩
 func validateAssurancesSignature(validators safrole.ValidatorsData, header block.Header, assurances block.AssurancesExtrinsic) error {
 	for _, assurance := range assurances {
-		if int(assurance.ValidatorIndex) >= common.NumberOfValidators || validators[assurance.ValidatorIndex] == nil {
```
greywolve, line 2174: nit: might be useful to have a different error here when the validator is present but empty, ie they are misbehaving, rather than just "bad validator index"

greywolve (approved):

internal/statetransition/state_transition.go
bamzedev: Makes sense, somewhat beyond the scope of this tho.

PR #327, opened 2025-05-20
Deserialize storage keys.

greywolve:
- Change the storage key used to a state key instead. This enables us to be able to deserialize the key later since serialization is lossy. We don't need the original key since the PVM code that will make host calls always knows the key it used. How we store that data and the key we use to store it under is merely an implementation detail. As long as we can set and get that data, it doesn't matter.
- Add storage key deserialization which is effectively just taking the serialized storage state key and it's value and adding it to the appropriate services storage map. Serialization is now just as trivial.
- Update host calls to take the original key and produce the storage state key when writing/reading storage keys.


pantrif (dismissed):

internal/state/serialization/deserialization.go
```diff
@@ -126,11 +134,42 @@ func deserializeService(state *state.State, sk statekey.StateKey, encodedValue [
...
-// DeserializePreimageLookup deserializes a preimage lookup from the given state key and encoded value.
+// deserializeStorage deserializes a storage account from the given state key and encoded value.
+func deserializeStorage(state *state.State, sk statekey.StateKey, encodedValue []byte) error {
+	if !sk.IsStorageKey() {
+		return fmt.Errorf("deserialize storage: expected storage key, got '%x'", sk[:])
+	}
+
+	serviceId, _, err := sk.ExtractServiceIDHash()
+	if err != nil {
+		return err
```
bamzedev, line 145: Wrap like the rest?

internal/state/serialization/statekey/statekey.go
```diff
@@ -170,6 +170,17 @@ func (s StateKey) IsServiceKey() bool {
 	return true
 }
 
+// Checks if the given state key is a storage key of the format: [n0, 0xFE, n1, 0xFF, n2, 0xFF, n3, 0xFF, h4, h5,...]
+// Where n is the service ID (uint32) little endian encoded, and h is the hash component.
+func (s StateKey) IsStorageKey() bool {
+	// The preimage lookup keys hash component starts with max(uint32) - 1
+	// little endian encoded, which is 0xFEFFFFFF. This is interleaved with the
```
bamzedev, line 177: Does the GP say that or jam encoded?

greywolve: Same thing in this case.

tests/integration/vectors_community/json/statedump.json
```diff
@@ -842,7 +842,7 @@
                     }
                 ],
                 "storage": {
-                    "0x81e47a19e6b29b0a65b9591762ce5143ed30d0261e5d24a3201752506b20f15c": "0xe9ac0c50"
+                    "0x81e47a19e6b29b0a65b9591762ce5143ed30d0261e5d24a3201752506b20f1": "0xe9ac0c50"
```
greywolve, line 845: This state dump was taken before the change to 31 byte state keys, in case you're wondering. :)

internal/state/serialization/statekey/statekey.go
greywolve: JAM encoded, but I'm not sure that's as clear as saying little endian encoded.

bamzedev: Not very likely, but possible the codec changes. We should not go around the repo chasing deviations like this one.

bamzedev (approved):

Issue #165, opened 2024-12-05
Safrole Protocols


bamzedev:
CE 131/132: Safrole ticket distribution

PR #334, opened 2025-05-23
Add safrole protocols

bamzedev:
https://github.com/eigerco/strawberry/issues/165
CE131/132 are identical aside from using different type of stream. Not everything is implemented, there are timings in which tickets needs to be send, this is a TODO.

The safrole happens in 2 steps:
1. Everyone sends their tickets to a calculated proxy vlaidator.
2. Proxy validator sends the tickets he received to all validators.

### DIFF
--- a/pkg/network/handlers/safrole_ticket_submit.go
+++ b/pkg/network/handlers/safrole_ticket_submit.go
@@ -1,0 +1,228 @@
+package handlers
+
+import (
+	"context"
+	"crypto/ed25519"
+	"fmt"
+
+	"github.com/eigerco/strawberry/internal/block"
+	"github.com/eigerco/strawberry/internal/jamtime"
+	"github.com/eigerco/strawberry/internal/state"
+	"github.com/eigerco/strawberry/internal/store"
+	"github.com/eigerco/strawberry/internal/validator"
+	"github.com/eigerco/strawberry/pkg/serialization/codec/jam"
+	"github.com/quic-go/quic-go"
+)
+
+// SafroleTicketSubmitRequestHandler handles CE 131 Safrole ticket submission protocol.
+// This is the first step in the two-phase ticket distribution process where
+// a generating validator sends their ticket to a deterministically-selected proxy validator.
+//
+// Protocol Flow (CE 131):
+// 1. Validator generates Safrole ticket for next epoch
+// 2. Validator determines proxy using: last 4 bytes of VRF output % validator_count
+// 3. Validator sends ticket to proxy validator (this handler receives it)
+// 4. Proxy validates ticket and stores it for later distribution (CE 132)
+type SafroleTicketSubmitRequestHandler struct {
+	safroleTicketHandler
+}
+
+// SafroleTicketDistributionRequestHandler handles CE 132 Safrole ticket distribution protocol.
+// This is the second step where the proxy validator broadcasts the received ticket
+// to all current validators for use in block production.
+//
+// Protocol Flow (CE 132):
+// 1. Proxy validator receives ticket via CE 131 (handled by submit handler above)
+// 2. After timing delay (3 minutes), proxy broadcasts to ALL current validators
+// 3. Current validators receive ticket via this handler
+// 4. Validators store ticket for potential use in block sealing
+type SafroleTicketDistributionRequestHandler struct {
+	safroleTicketHandler
+}
+
+// safroleTicketHandler contains the common logic for both CE 131 and CE 132 protocols.
+// Both protocols use identical wire format but different StreamKind and serve different roles in the distribution process.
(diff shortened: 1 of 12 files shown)


pkg/network/handlers/safrole_ticket_submit.go
```diff
@@ -0,0 +1,228 @@
...
+		return fmt.Errorf("unmarshal message: %w", err)
+	}
+
+	// Get the ring commitment for the next epoch's validators
+	// This is used to verify the RingVRF proof in the ticket
+	// TODO: Deal with epoch change.
+	ringCommitment, err := handler.state.ValidatorState.SafroleState.NextValidators.RingCommitment()
+	if err != nil {
+		return fmt.Errorf("get ring commitment: %w", err)
+	}
```
greywolve, line 150:
```suggestion
	ringCommitment, err := handler.state.ValidatorState.SafroleState.RingCommitment
```

No need to recompute this, just take the one that's already there.

pkg/network/handlers/safrole_ticket_submit.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// For CE 131: Verify this node is the designated proxy for this ticket
+	// For CE 132: This check should be skipped as all validators should accept broadcast tickets. requireProxyValidation should be true.
+	if requireProxyValidation && !handler.validatorManager.IsProxyValidatorFor(hash) {
+		return fmt.Errorf("not proxy validator for hash %v", hash)
+	}
+
+	// Calculate the next epoch when this ticket will be used
+	nextEpoch, err := jamtime.CurrentEpoch().NextEpoch()
```
greywolve, line 166: A bit confused here. Tickets should be used in the current epoch, not the next epoch. They are used in the current epoch to produce new sealing keys which get used in the next epoch, but these will already be in the state by that point.

greywolve: Tickets are stored in the accumulator, and then become sealing keys, so there's no much reason to store them beyond a 'mempool' for the current epoch.

pkg/network/handlers/safrole_ticket_submit.go
```diff
@@ -0,0 +1,228 @@
...
+// <-- FIN
+func (r *SafroleTicketSubmiter) Submit(ctx context.Context, stream quic.Stream, ticketProof block.TicketProof) error {
+	nextEpoch, err := jamtime.CurrentEpoch().NextEpoch()
+	if err != nil {
+		return fmt.Errorf("get next epoch: %w", err)
+	}
+
+	// Create the message with epoch index and ticket proof
+	msg := safroleTicketSubmitMessage{
+		EpochIndex:  nextEpoch,   // When the ticket will be used
```
greywolve, line 207: Think this should also be current epoch not next.

pkg/network/node/node.go
```diff
@@ -597,6 +608,86 @@ func (n *Node) RequestState(ctx context.Context, headerHash crypto.Hash, keyStar
...
+// - CE 131: Generating validator -> Proxy validator (1-to-1)
+// - CE 132: Proxy validator -> All current validators (1-to-many broadcast)
+//
+// Protocol Details:
+// - Both use the same message format: Epoch Index + Ticket data
+// - Epoch Index identifies which epoch the ticket will be used for block production
+// - Simple request-response: sender transmits data, receiver acknowledges with FIN
+func (n *Node) sendTicket(ctx context.Context, streamKind protocol.StreamKind, ticket block.TicketProof, peerKey ed25519.PublicKey) error {
+	n.peersLock.RLock()
+	defer n.peersLock.RUnlock()
```
danielvladco, line 672: The peers set will be blocked for the entire time the stream is opened. This could potentially slow down the node a lot, can we just use lock/unlock to get the peer and continue working with it without blocking for the entirety of the execution?

pkg/network/handlers/safrole_ticket_submit.go
bamzedev: I don't know enough about how they are used safrolewise, but this is what the network protocol does. You send them in the previous epoch, the proxy keeps them and after the epoch change distributes them. The timing of distributing is a TODO atm.

pkg/network/handlers/safrole_ticket_submit.go
bamzedev: I do, it's for next validators, not current ones.

bamzedev: You are right.

pkg/network/handlers/safrole_ticket_submit.go
bamzedev: You are right.

Issue #336, opened 2025-05-26
Wrong JAM epoch time


toncek345:
https://graypaper.fluffylabs.dev/#/9a08063/09d60109da01?v=0.6.6
https://github.com/eigerco/strawberry/blob/main/internal/jamtime/jamtime.go#L10-L12

The start of JAM epoch is - 12:00 UTC January 1, 2025

---
Another thing is changing the `now` function to return time in UTC. Currently it will return it based on timezone the machine is set to.

https://github.com/eigerco/strawberry/blob/main/internal/jamtime/jamtime.go#L8
```go
var now = func() time.Time {
	return time.Now().UTC()
}
```

PR #337, opened 2025-05-27
Fix jamtime to always use UTC

bamzedev:
This PR ensures that all time operations in the JAM time package consistently use UTC. 
Updates the start of JAM according to the update in the Graypaper.
Removed `//go:build !integration` from the tests as there is no need for it.

Fixes https://github.com/eigerco/strawberry/issues/336.


internal/jamtime/jamtime.go
```diff
@@ -26,14 +28,15 @@ type JamTime struct {
 
 // Now returns the current time as a JamTime
 func Now() JamTime {
-	t := now()
+	t := now().UTC()
```
toncek345, line 31: `.UTC()` is not necesary here because it is already in the function pointer on line 9.

greywolve (approved):

internal/jamtime/jamtime.go
bamzedev: It is covered https://github.com/eigerco/strawberry/pull/337/files#diff-2bbd57291ebcb75adb381153f01b56f8bb2e2415f28e2db5fe4369e3dd201522R381. Anyway, you see this as an issue?

toncek345 (approved):

internal/jamtime/jamtime.go
toncek345: It's fine, resolved-closed

greywolve (approved):

PR #342, opened 2025-05-27
Simulation test helper that generates valid sealed blocks.

greywolve:
Add a test helper which generates valid sealed blocks based on a pre state and validator keys and also prints a diff between the resulting post state and pre state. Useful for iterating on simulation tests.

Also:
- Move other simulation test helpers to utils.go.
- Improve NextSafroleState, rather just use the existing STF function rather then reinventing the wheel.
- Fix a bug in the safrole integration test where we weren't correctly parsing ed25519 keys in the epoch marker. I noticed that during the safrole simulation run that the ed25519 keys were blank.


tests/simulation/simulation_helper_test.go
```diff
@@ -0,0 +1,103 @@
+//go:build integration && generate
```
bamzedev, line 1: What is the use going to be here? When do we run this?

tests/simulation/simulation_helper_test.go
```diff
@@ -0,0 +1,103 @@
...
+	require.NoError(t, err)
+
+	nextTimeslot := templateBlock.Header.TimeSlotIndex
+
+	_, slotLeaderKey, err := FindSlotLeader(
+		nextTimeslot,
+		currentState,
+		keys,
+	)
+
```
bamzedev, line 66:
```suggestion
require.NoError(t, err)
```

tests/simulation/simulation_helper_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/eigerco/strawberry/internal/state"
+	"github.com/eigerco/strawberry/internal/statetransition"
+	"github.com/eigerco/strawberry/internal/store"
+	jsonutils "github.com/eigerco/strawberry/internal/testutils/json"
+	"github.com/eigerco/strawberry/pkg/db/pebble"
+	"github.com/pmezard/go-difflib/difflib"
+	"github.com/stretchr/testify/require"
+)
+
+// TestGenerateGenesisState is used to create valid blocks for simulation test
```
bamzedev, line 20: Test name is TestBlockGenerator

tests/simulation/simulation_helper_test.go
```diff
@@ -0,0 +1,103 @@
...
+		templateBlock.Header.ParentHash,
+		currentState,
+		trieDB,
+		slotLeaderKey,
+		templateBlock.Extrinsic,
+	)
+	require.NoError(t, err)
+
+	// Dump the valid selaed block
+	os.WriteFile("block_out.json", []byte(jsonutils.DumpBlockSnapshot(newBlock)), 0644)
```
bamzedev, line 78:
```suggestion
	err = os.WriteFile("block_out.json", []byte(jsonutils.DumpBlockSnapshot(newBlock)), 0644)
require.NoError(t, err)
```

bamzedev:
One more thing, the way this dumps `jsons`, we will be committing them unwillingly.

greywolve:
> One more thing, the way this dumps `jsons`, we will be committing them unwillingly.

Do you have any suggestions? Maybe this should be behind a flag to dump the json?

bamzedev:
Specific output folder that we add to ignore.

bamzedev (approved):

PR #348, opened 2025-05-30
Update codec test vectors

pantrif:


pkg/serialization/codec/jam/encode.go
```diff
@@ -402,6 +402,19 @@ func (bw *byteWriter) encodeStruct(in interface{}) error {
 				}
 				continue
 			}
+			// Handle compact encoding for unsigned integers if specified via struct tag
+			if encodingType, found := tagValues["encoding"]; found && encodingType == "compact" {
```
bamzedev, line 406: Check for conflicting tags maybe? `length` and `compact` ?

tests/integration/vectors/extrinsic_hash/preimages_guarantees.bin
bamzedev, line 1: Why the `bin` files?

pkg/serialization/codec/jam/encode_decode_jam_test.go
```diff
@@ -158,6 +158,38 @@ func TestLengthTag(t *testing.T) {
...
+	require.Equal(t, []byte{10, 0, 0, 0}, marshaledData)
+
+	var noTagUnmarshaled NoTag
+	err = jam.Unmarshal(marshaledData, &noTagUnmarshaled)
+	require.NoError(t, err)
+	assert.Equal(t, noTag, noTagUnmarshaled)
+
+	// simple struct with compact tag
+	type WithTag struct {
+		Uint32 uint32 `jam:"encoding=compact"`
```
bamzedev, line 179: We should have test for each type.

pkg/serialization/codec/jam/encode_decode_jam_test.go
```diff
@@ -158,6 +158,38 @@ func TestLengthTag(t *testing.T) {
 	assert.Equal(t, original, unmarshaled)
 }
 
+func TestCompactTag(t *testing.T) {
```
bamzedev, line 161: Given how central the codec is, we should not shy away from tests. Check error cases?

greywolve:
I think we should probably also switch [core and service stats](https://github.com/eigerco/strawberry/blob/update-codec-reports-test-vectors/internal/validator/validators.go#L37-L68) to the new tag.

bamzedev (approved):

greywolve (approved):

PR #351, opened 2025-06-02
feat: reconstruct segment shards from assurers

danielvladco:


bamzedev:
It's from a prev pr, but would you mind adding a check for the err [here](https://github.com/eigerco/strawberry/blob/main/pkg/network/network_test.go#L1252) ?

internal/d3l/segments_fetcher_test.go
```diff
@@ -0,0 +1,143 @@
...
+	return args.Get(0).([]*peer.Peer)
+}
+
+func (a *assurerClientMock) SegmentShardRequestSend(ctx context.Context, peerKey ed25519.PublicKey, erasureRoot crypto.Hash, shardIndex uint16, segmentIndexes []uint16) (segmentShards [][]byte, err error) {
+	args := a.MethodCalled("SegmentShardRequestSend", ctx, peerKey, erasureRoot, shardIndex, segmentIndexes)
+	return args.Get(0).([][]byte), args.Error(1)
+}
+
+func TestNewSegmentsFetcher(t *testing.T) {
+	assurers := make([]*peer.Peer, common.NumberOfValidators)
```
pantrif, line 36: consider adding the build tag for tiny configuration to avoid slowing down the pipeline

pkg/network/handlers/shard_distribution.go
```diff
@@ -14,7 +15,7 @@ import (
 	"github.com/eigerco/strawberry/pkg/serialization/codec/jam"
 )
 
-const SegmentShardLength = 12
+const SegmentShardLength = 4104 / common.ErasureCodingOriginalShards
```
pantrif, line 18:
```suggestion
const SegmentShardSize = 4104 / common.ErasureCodingOriginalShards
```

internal/d3l/segments_fetcher_test.go
danielvladco: I am not sure if we should do this, we already use tiny configuration for the e2e test, I think we should have at least a unit test that tests with the full configuration

greywolve (approved):

pantrif (approved):

bamzedev (approved):

PR #363, opened 2025-06-05
feat: log host call

danielvladco:


greywolve:
> > Shouldn't authorization and refine also have this host call?
> 
> Yes, but now we are working on M1 so I thought not to waste time and touch those parts of the code, also refine has to log the core, and from the currently I think we are not passing it down to the refine function, so this would require more refactoring, so for now better to ignore it

Could we add a TODO somewhere to remember to add it? maybe in each section?

PR #365, opened 2025-06-05
Available reports should be values not pointers

greywolve:
Available reports should be []block.WorkReport, not []*block.WorkReport.

Also pass the correct available reports to accumulate in UpdateState.


tests/integration/assurances_test.go
```diff
@@ -127,15 +128,15 @@ func mapSlice[T1, T2 any](s1 []T1, fn func(T1) T2) (s2 []T2) {
 	return
 }
 
-func mapReport(r *Report) *block.WorkReport {
+func mapReport(r *Report) block.WorkReport {
 	if r == nil {
```
danielvladco, line 132: use `require.NotNil(t, r)`

internal/statetransition/state_transition.go
```diff
@@ -65,7 +65,7 @@ func UpdateState(s *state.State, newBlock block.Block, chain *store.Chain) error
 		return err
 	}
 
-	intermediateCoreAssignments, _, err = CalculateIntermediateCoreFromAssurances(s.ValidatorState.CurrentValidators, intermediateCoreAssignments, newBlock.Header, newBlock.Extrinsic.EA)
+	intermediateCoreAssignments, availableReports, err := CalculateIntermediateCoreFromAssurances(s.ValidatorState.CurrentValidators, intermediateCoreAssignments, newBlock.Header, newBlock.Extrinsic.EA)
```
danielvladco, line 68: This is the dependency graph according to GP
```
ρ‡ ≺ (EA, ρ†)(4.13)
ρ′ ≺ (EG, ρ‡, κ, τ ′)(4.14)
W∗ ≺ (EA, ρ′)(4.15)
```

however the work reports here are the WR of the intermediate core assignments (`ρ‡`) not the new core assignments `ρ'` so we should not use these work reports because most likely they are wrong, we should refactor `CalculateIntermediateCoreFromAssurances` altogether because it is no longer up to date with the GP

danielvladco: but also eq. 11.16 uses `ρ†` so it's neither `ρ‡` nor `ρ'`. I didn't look in depth what is the difference between these three, maybe they are equivalent in regards to work reports. This needs to be checked properly and if the dependency graph above is wrong we should create a PR to fix the GP

greywolve: I have a [PR](https://github.com/gavofyork/graypaper/pull/416) up to fix the GP, because yeah, that's wrong.

11.16-11.17 together produce `ρ‡` `W` are the reports that are available and are removed from  `ρ†` to produce `ρ‡`

greywolve: You can't get `W` from `p'`,  because `p'` only contains reports that are waiting to be made available.

greywolve:
We can get to this later. Closing for now.

PR #385, opened 2025-06-25
Block header verification

greywolve:
Add two functions which together do comprehensive block header verification:
- VerifyBlockHeaderBasic, which validates header fields that only
  require the prior state.
- VerifyBlockHeaderSafrole which validates header fields which rely on
  an updated SAFROLE state to proceed.

Right now this is tested as part of safrole simluation, but more tests
to follow in a future PR.

Also:
- Fix up simulation tests which needed to have updated extrinsic hashes
  and parent state roots.
- Skip guarantees simulation test for now, updating the hashes didn't
  help because we also need to generate new signatures for the work
report after updating the state root it should point to.
- Move safrole related STF functions out into it's own file. A separate
  package would have been better, but right now that's not possible
until we structure our packages a little differently to avoid circular
deps.
- Add NextSafroleState function which can be used outside of UpdateState
  to determine the next Safrole state. This function has been moved from
simulation and updated to also potentially handle incoming disputes.
It's used only in simulation for now.
- Offenders marker will be handled later, added a TODO.

### DIFF
--- a/internal/statetransition/block.go
+++ b/internal/statetransition/block.go
@@ -1,0 +1,125 @@
+package statetransition
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/eigerco/strawberry/internal/block"
+	"github.com/eigerco/strawberry/internal/state"
+	"github.com/eigerco/strawberry/internal/state/merkle"
+	"github.com/eigerco/strawberry/internal/store"
+	"github.com/eigerco/strawberry/internal/validator"
+)
+
+// VerifyBlockHeaderBasic verifies header fields which are verifiable with just
+// the prior state:
+// - The timeslot
+// - The prior state root
+// - The extrinsic hash
+// The trie database must be passed in produce the merkle root of the state.
+// GP v0.7.0
+func VerifyBlockHeaderBasic(priorState *state.State, block block.Block, trie *store.Trie) error {
+	// Timeslot validation. Equation 5.7
+	// P(H)_T < H_T ∧ H_T ⋅ P ≤ T
+	if block.Header.TimeSlotIndex.IsInFuture() {
+		return errors.New("timeslot is in the future")
+	}
+	if priorState.TimeslotIndex >= block.Header.TimeSlotIndex {
+		return errors.New("timeslot must be greater than the prior state's timeslot")
+	}
+
+	// State root validation. Equation 5.8
+	//  H_R ≡ M_σ(σ)
+	expectedPriorStateRoot, err := merkle.MerklizeState(*priorState, trie)
+	if err != nil {
+		return fmt.Errorf("failed to merklize state: %w", err)
+	}
+	if block.Header.PriorStateRoot != expectedPriorStateRoot {
+		return fmt.Errorf("invalid prior state root: %x", expectedPriorStateRoot)
+	}
+
+	// Check extrinsic hash. Equation 5.4
+	// H_X ≡ H(E(H#(a)))
+	expectedExtrinsicsHash, err := block.Extrinsic.Hash()
+	if err != nil {
+		return fmt.Errorf("failed to hash extrinsics: %w", err)
+	}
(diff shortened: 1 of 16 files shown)


internal/statetransition/block.go
```diff
@@ -0,0 +1,108 @@
...
+	offenders := make([]ed25519.PublicKey, 0, len(block.Extrinsic.ED.Culprits)+len(block.Extrinsic.ED.Faults))
+	for _, c := range block.Extrinsic.ED.Culprits {
+		offenders = append(offenders, c.ValidatorEd25519PublicKey)
+	}
+	for _, f := range block.Extrinsic.ED.Faults {
+		offenders = append(offenders, f.ValidatorEd25519PublicKey)
+	}
+
+	for i, o := range offenders {
+		if !block.Header.OffendersMarkers[i].Equal(o) {
```
bamzedev, line 53: I believe we are not populating this atm. I told you I am not updating this for now, why are you adding a check for it?

internal/statetransition/state_transition.go
```diff
@@ -35,10 +34,6 @@ import (
 //
 //	it might be worth making State immutable and make it so that UpdateState returns a new State with all the updated fields
 func UpdateState(s *state.State, newBlock block.Block, chain *store.Chain) error {
-	if newBlock.Header.TimeSlotIndex.IsInFuture() {
```
bamzedev, line 38: So this check is now gone. Seems like a bug.

internal/statetransition/block.go
```diff
@@ -0,0 +1,108 @@
...
+	// Check extrinsic hash.
+	expectedExtrinsicsHash, err := block.Extrinsic.Hash()
+	if err != nil {
+		return fmt.Errorf("failed to hash extrinsics: %w", err)
+	}
+	if block.Header.ExtrinsicHash != expectedExtrinsicsHash {
+		return errors.New("invalid extrinsics hash")
+	}
+
+	// Check offenders marker.
```
bamzedev, line 43: I don't think we should go through the extrinsics manually here. The correct list is produced in `CalculateNewJudgements`, you can take that as an argument. Also the length calculation is wrong.

internal/statetransition/block.go
bamzedev, line 1: Ideally we would have some equations to go with all these checks.

internal/statetransition/block.go
greywolve: I guess the intention here is to be able to use this to verify blocks in general, not just for use with UpdateState. (ie nodes would use this to reject blocks early) That would mean we'd have to also do `CalculateNewJudgements` in here too, which I suppose is doable. I'd prefer if this was self contained, so I'm not sure an argument with offenders is the way to go.

internal/statetransition/block.go
greywolve: We are populating this in simulation, or at least there's code to do it, might be wrong: https://github.com/eigerco/strawberry/blob/main/tests/simulation/utils.go#L136

If you'd prefer to add this part of the validation yourself, I could remove anything to do with the offenders field and leave that as a TODO for later.

internal/statetransition/state_transition.go
greywolve: It's moved to the VerifyBlockHeader, which should be called before passing the block in. What would you suggest?

internal/statetransition/block.go
bamzedev: The state transition does not bother with it as its not needed atm. I prefer not having to deal with optional stuff.

internal/statetransition/block.go
greywolve: Ie as soon as you get a block sent to you, you can check if it's valid

internal/statetransition/state_transition.go
bamzedev: Call it in here at the start.

PR #391, opened 2025-07-02
Safrole equation comment updates for v0.7.0

greywolve:


greywolve:
> Would you mind putting actual equation and version? In the places where it even misses both makes it almost useless. Eq. 6.15 5 versions from would prolly mean nothing.

I added the version number to the top of each file. (maybe you missed that) I think these should probably be updated together anyway. This way I can just mention the version once and not repeat it with every equation.

bamzedev:
> > Would you mind putting actual equation and version? In the places where it even misses both makes it almost useless. Eq. 6.15 5 versions from would prolly mean nothing.
> 
> I added the version number to the top of each file. (maybe you missed that) I think these should probably be updated together anyway. This way I can just mention the version once and not repeat it with every equation.

You believe that every version you are going to update each equation number if they have changed?

greywolve:
> > > Would you mind putting actual equation and version? In the places where it even misses both makes it almost useless. Eq. 6.15 5 versions from would prolly mean nothing.
> > 
> > 
> > I added the version number to the top of each file. (maybe you missed that) I think these should probably be updated together anyway. This way I can just mention the version once and not repeat it with every equation.
> 
> You believe that every version you are going to update each equation number if they have changed?

For this section I don't suspect there will be much changes going forward. Equation numbers haven't really changed since v0.5.4. But if you prefer adding version numbers to each equation despite that, I'm happy to.

bamzedev:
> > > > Would you mind putting actual equation and version? In the places where it even misses both makes it almost useless. Eq. 6.15 5 versions from would prolly mean nothing.
> > > 
> > > 
> > > I added the version number to the top of each file. (maybe you missed that) I think these should probably be updated together anyway. This way I can just mention the version once and not repeat it with every equation.
> > 
> > 
> > You believe that every version you are going to update each equation number if they have changed?
> 
> For this section I don't suspect there will be much changes going forward. Equation numbers haven't really changed since v0.5.4. But if you prefer adding version numbers to each equation despite that, I'm happy to.

Lets say [this](https://github.com/eigerco/strawberry/pull/391/files#diff-3d9ee3bf0e5abd55e804ba2cc26515fc89f92d6b7b8f87739b836cbd8a321297R275) becomes C23 in next version. What are you going to do

> > > > Would you mind putting actual equation and version? In the places where it even misses both makes it almost useless. Eq. 6.15 5 versions from would prolly mean nothing.
> > > 
> > > 
> > > I added the version number to the top of each file. (maybe you missed that) I think these should probably be updated together anyway. This way I can just mention the version once and not repeat it with every equation.
> > 
> > 
> > You believe that every version you are going to update each equation number if they have changed?
> 
> For this section I don't suspect there will be much changes going forward. Equation numbers haven't really changed since v0.5.4. But if you prefer adding version numbers to each equation despite that, I'm happy to.

It makes sense when an equation change you change that function and you put a version on it, instead of updating the whole file. What happens in `state transition` in such case? It's just clean code comments, I don't get what we save by skipping adding useful comments, it takes minutes. Also adding `eq 6.15-6.20` on all the functions is kind of pointless as well.

PR #404, opened 2025-08-12
Add fallback traces for v0.6.7

greywolve:
Also fix a bug with setting the most recently accumulated timeslot field in services.


internal/statetransition/state_transition.go
```diff
@@ -627,7 +627,7 @@ func CalculateWorkReportsAndAccumulate(header *block.Header, currentState *state
 		//      ⎧ a except a′_a = τ′ if s ∈ K(S)
 		// a′ = ⎨
 		//      ⎩ a otherwise
-		if _, ok := accumulationStats[serviceId]; !ok {
+		if _, ok := accumulationStats[serviceId]; ok {
```
greywolve, line 630: @danielvladco I think you need to double check this fix. :) Traces were failing because this was incorrectly setting the most recently accumulated timeslot before. In the traces this just stays 0.

danielvladco: it's perfect, I also made this fix on my branch 👍

PR #419, opened 2025-09-29
Update fuzzer to v1

bamzedev:
Updating fuzzer to v1 - adding the error type, forks, ancestry option.


internal/state/serialization/serialization.go
```diff
@@ -161,3 +161,17 @@ func sortByteSlicesCopy(slice interface{}) interface{} {
 		panic("unsupported type for sorting")
 	}
 }
+
+func CloneState(in state.State) (state.State, error) {
```
greywolve, line 165: This is probably going to much slower, but should be good enough for now.

pkg/conformance/node_test.go
```diff
@@ -242,71 +247,6 @@ func TestMessage(t *testing.T) {
 
 }
 
-func TestHandshake(t *testing.T) {
```
greywolve, line 245: Removing this because the handshake is already being tested?

pkg/conformance/node.go
```diff
@@ -104,10 +108,20 @@ func (n *Node) handleConnection(conn net.Conn) {
 
 		responseMsg, err := n.messageHandler(msg)
 		if err != nil {
-			log.Printf("error handling message: %v", err)
-			return
+			if strings.Contains(err.Error(), "preimage unneeded") {
```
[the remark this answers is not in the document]
bamzedev: Yeah, once we know all the errors. Need to pass all traces.

internal/state/serialization/serialization.go
bamzedev: Yeah, good for now.

